PR #1, opened 2019-11-14
PIC-37 Adding DynamoDB

markberridge:


helm_deploy/court-case-service/templates/_envs.tpl
```diff
@@ -19,4 +19,28 @@ env:
...
+        name: court-case-dynamodb-output
+        key: access_key_id
+
+  - name: DYNAMO_SECRET_KEY
+    valueFrom:
+      secretKeyRef:
+        name: court-case-dynamodb-output
+        key: secret_access_key
+
+  - name:
```
SWhitlam: you're missing the Env Var name here

src/main/java/uk/gov/justice/probation/courtcaseservice/CreateTableCommandLineRunner.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.stereotype.Component;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.awssdk.services.dynamodb.model.*;
+import uk.gov.justice.probation.courtcaseservice.config.DynamoDbConfiguration;
+
+@Component
+@Profile("localDynamo")
+public class CreateTableCommandLineRunner implements CommandLineRunner {
+
+    private final Logger log = LoggerFactory.getLogger(CreateTableCommandLineRunner.class);
```
SWhitlam: You could use the lombok annotation here, but not sure if this is just a test class with it being a CLI runner

src/main/java/uk/gov/justice/probation/courtcaseservice/config/CourtCaseServiceProperties.java
```diff
@@ -1,6 +1,5 @@
-package uk.gov.justice.probation.courtlistservice.config;
+package uk.gov.justice.probation.courtcaseservice.config;
```
SWhitlam: Empty class

helm_deploy/court-case-service/templates/_envs.tpl
markberridge: We don't need the ARN so I have removed it

PR #2, opened 2019-11-26
CORS support

theDustRoom:
Allows Cross-origin resource sharingCross-origin resource sharing on /ping endpoint so that healthcheck can be hit from Frontend code (such as build radiator)


markberridge:
LGTM

PR #9, opened 2019-12-09
Pic 182 add flyway

katie-cawthorne:
- Added configuration for Flyway database migration

- Added SQL scripts 

- Created Spring Data entities


build.gradle
```diff
@@ -6,6 +6,7 @@ plugins {
     id 'org.springframework.boot' version '2.1.9.RELEASE'
     id 'io.spring.dependency-management' version '1.0.8.RELEASE'
     id 'com.github.ben-manes.versions' version '0.26.0'
+    id "org.flywaydb.flyway" version "6.1.0"
```
SWhitlam: Not sure if you need this plugin

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CourtCaseEntity.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Column(name = "COURT_ROOM")
+    private String courtRoom;
+
+    @Column(name = "SESSION_START_DATE")
+    private LocalDateTime sessionStartDate;
+
+    @Type(type = "jsonb")
+    @Column(columnDefinition = "jsonb", name = "DATA")
+    private CourtCasePropertiesEntity data;
```
SWhitlam: This is a good way of using JSONB but the downside is you're restricted to a specific schema for the JSONB field. If you want more flexibility you may want to have a normal JSONB field and manually parse the values to/from a String.

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CourtEntity.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.justice.probation.courtcaseservice.jpa.entity;
```
SWhitlam: Probably not important at this point but you will probably want to store an external Court ID here from Delius or the Court service?

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtCaseRepository.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+import uk.gov.justice.probation.courtcaseservice.jpa.entity.CourtCaseEntity;
+
+import java.util.Optional;
+
+@Repository
+public interface CourtCaseRepository extends CrudRepository<CourtCaseEntity, Long> {
+
+    Optional<CourtCaseEntity> findByCaseId(long caseId);
```
SWhitlam: You may get this method for free from Spring Data.

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtRepository.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+import uk.gov.justice.probation.courtcaseservice.jpa.entity.CourtEntity;
+
+import java.util.Optional;
+
+@Repository
+public interface CourtRepository extends CrudRepository<CourtEntity, Long> {
+
+    Optional<CourtEntity> findById(long Id);
```
SWhitlam: As above this might be already implemented for you by Spring.

src/main/resources/db/migration/V1__create_court_case.sql
```diff
@@ -0,0 +1,25 @@
...
+    NAME                                   TEXT         NOT NULL
+);
+
+
+DROP TABLE IF EXISTS COURT_CASE;
+
+CREATE TABLE IF NOT EXISTS COURT_CASE
+(
+  ID                                     SERIAL       PRIMARY KEY,
+  CASE_ID                                BIGINT       NOT NULL,
```
SWhitlam: Where does this value from from? Is it generated?

build.gradle
katie-cawthorne: It's a flyway gradle plugin I found with useful tasks that you can run, such as `gradle flywayInfo` that prints the details and status information about all the migrations. 
I know you can do this from command line so don't necessarily need.

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CourtEntity.java
katie-cawthorne: It's just a starter for 10, we'll add that in when we know more about the api.

PR #16, opened 2020-01-15
Pic 158 identify defendants current to probation

katie-cawthorne:
Added more configuration for flyway


src/main/resources/db/migration/courtcase/V1__initial_setup.sql
```diff
@@ -0,0 +1,24 @@
+DROP TABLE IF EXISTS COURT;
```
jhevans: @katie-cawthorne unfortunately the changes to existing migrations have caused some build issues (see https://dsdmoj.atlassian.net/browse/PIC-217) so I think these changes are likely to cause the same problem.  Any changes to the database need to be done in sequence or it will confuse Flyway in the deployed envs 😕.

PR #17, opened 2020-01-15
Pic 183 create case

theDustRoom:


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CourtEntity.java
```diff
@@ -3,15 +3,16 @@
...
 import javax.persistence.*;
 import java.io.Serializable;
 
-
 @Entity
 @AllArgsConstructor
 @NoArgsConstructor
 @Data
 @Table(name = "COURT")
+@Cacheable
```
jhevans: Will this annotation work on its own? From what I can tell this requires a bit of [additional configuration](https://www.baeldung.com/hibernate-second-level-cache) to set up the L2 cache. Have we done this? If not I'd argue it might be worth leaving this until we actually need it for performance reasons. If we introduce caching we probably ought test it specifically to make sure it's applied correctly.
```suggestion
```

theDustRoom: I read that with javax.persistence.Cacheable it should store to L2 cache but now after having read more there’s conflicting information, some suggest it’s there out of the box and some state otherwise. I don’t know anywhere near enough about JPA and Spring.

Mark did comment on the card that Court entity should be stored in L2 cache.

Either way I guess we do need tests around it to ensure it’s functioning as required.

I’ll look into it more this morning.

theDustRoom: So after reading up a little and then speaking with Mark we're going to look into cacheing at a later date and put a card on the backlog for this.

As such; I've removed the @Cacheable annotation.

PR #18, opened 2020-01-17
Pic 213 get case list endpoint

jhevans:
Implemented endpoint to bring back all cases with session time on a given date


src/test/resources/before-test.sql
```diff
@@ -4,5 +4,12 @@ DELETE FROM courtcaseservicetest.court WHERE true;
...
 
+-- See CourtCaseControllerTest.shouldGetCaseListWhenCasesExist()
+-- These records are used to test edge cases when returning court case list for a given date (midnight to 1 second before midnight the next day)
```
katie-cawthorne: Sorry due to my recent database changes, this will need amending.

jhevans: No problem, updated and all tests passing

PR #24, opened 2020-02-06
PIC-243, PIC-237: Add smoke tests and resolve build issues

jhevans:
Introduced some smoke tests which can be run against all environments and updated Pipeline to add these as a step.

Also resolved an issue with intermittent build failures we've been having by centralising static config of RestAssured.


jhevans:
Having run the smoke tests I've discovered that we've got a fairly major issue in that we're being blocked by our ingress rules. I don't think this is going to be trivial to solve without making the rules overly permissive so I'm not going to merge this PR until we can work out an alternative

jhevans:
> Having run the smoke tests I've discovered that we've got a fairly major issue in that we're being blocked by our ingress rules. I don't think this is going to be trivial to solve without making the rules overly permissive so I'm not going to merge this PR until we can work out an alternative

Have just backed out the circleci config changes so we can merge the rest

PR #26, opened 2020-02-10
Pic 206 add previously known termination date

katie-cawthorne:
PIC-206:
- Added `previouslyKnownTerminationDate` column to `court_case` table

- Added V3 migration file for changes to the database

- In order to amend the seed data I created an undo file `U99__seed_data.sql` to undo migration `V99__seed_data.sql` and created a repeatable migration for the seed data `R__seed_data.sql` that isn't versioned and can be modified, so will remove any issues we have going forward when we need to amend the seed data.

- Added tests


src/main/resources/db/migration/development/U99__seed_data.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO court (id, name, court_code) VALUES (1142407, 'Sheffield Magistrates Court', 'SHF');
```
jhevans: An undo migration would normally undo the effects of the original migration so I'd expect this to be along the lines of `DELETE FROM court_case where * id in [blah....]`. Not sure I understand why we need it so I'll give you a shout when we're both in if that's cool.

PR #27, opened 2020-02-12
Pic 209 add suspended sentence order

katie-cawthorne:
- Added suspendedSentenceOrder field to entity

- Added tests

- Added new migration file for database change and updated seed file


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CourtCaseEntity.java
```diff
@@ -47,6 +48,12 @@
     @Column(name = "PROBATION_STATUS")
     private String probationStatus;
 
+    @Column(name = "PREVIOUSLY_KNOWN_TERMINATION_DATE")
+    private LocalDate previouslyKnownTerminationDate;
+
+    @Column(name = "SUSPENDED_SENTENCE_ORDER")
+    private Boolean suspendedSentenceOrder;
```
jhevans: This should probably default to `false`

PR #28, opened 2020-02-18
Pic 251 community api auth

jhevans:


src/main/resources/application-local.yml
```diff
@@ -0,0 +1,17 @@
...
+
+spring:
+  security:
+    oauth2:
+      client:
+        registration:
+          nomis-oauth-client:
+            client-name: nomis-oauth-client
+            client-id: community-api-client
+            client-secret: clientsecret
```
markberridge: can you change this to a placeholder value with the default value of `clientsecret` to prevent any temptation to hard code secrets in config

jhevans: Good call, will do

src/main/java/uk/gov/justice/probation/courtcaseservice/application/WebClientConfig.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientManager;
+import org.springframework.security.oauth2.client.web.reactive.function.client.ServletOAuth2AuthorizedClientExchangeFilterFunction;
+import org.springframework.web.reactive.function.client.WebClient;
+
+@Configuration
+@ConditionalOnProperty(name = "feature-flags.community-api-test-endpoint")
```
jhevans: @markberridge it occurred to me this was going to fall over as soon as we deployed to dev given we don't have valid credentials so wrapped this in a feature flag until such a time as we have them sorted in all envs. Also moved most of the config from application-local to application to prevent any missing properties causing it to fall over

PR #29, opened 2020-02-18
Flyway migration fixes:

jhevans:
- Bring all migrations already applied into db.migration.courtcase
- Introduce new migration to clear erroneous seed data from envs
- Clean up properties files
- Re add repeatable seed data migration


src/main/resources/db/migration/courtcase/V100__remove_seed_data.sql
```diff
@@ -0,0 +1,9 @@
+-- This migration undoes most of the effects (with the exception of adding the SHF court as this is needed) of
+-- V99__seed_data.sql which was inadvertently applied to prod.
+
+DELETE FROM court_case WHERE case_id='1168460';
```
chris-faulkner-mojdt: Presumably this is test-based data. So should be in src/test/resources so it doesn't get packaged with anything which would go near production.

jhevans: It's more sample data used for local deployments so you don't start with a completely empty DB and need to do a load of set up before you can actually use it. Good call though, will do as you suggest 👍

PR #37, opened 2020-03-06
Pic 285 get offender probation data

jhevans:
Adds a new offender endpoint which brings back data via two calls to the community api for  offender data and convictions by CRN


jhevans:
There's an intermittent build failure happening at the moment which I'm still investigating so this shouldn't be merged until that's been fixed. Otherwise it's code complete.

jhevans:
> There's an intermittent build failure happening at the moment which I'm still investigating so this shouldn't be merged until that's been fixed. Otherwise it's code complete.

Right, so I've managed to narrow this down and it's somehow being caused by oauth when running against Wiremock, I haven't managed to reproduce it running against the community api proper. For now I've disabled OAuth altogether to get this ticket through. We won't be able to authenticate with the community API in dev yet but there's work to be done before we can do that anyway so will raise a separate ticket.

PR #38, opened 2020-03-09
Fix dev deployment by providing default config value for community-api base url

jhevans:


jhevans:
@markberridge I've requested your review because you added the application-dev config file to .gitignore in the past. I assume that's no longer needed?

PR #41, opened 2020-03-10
PIC-312: Fix missing values in deployment

jhevans:


helm_deploy/values-preprod.yaml
```diff
@@ -27,3 +27,9 @@ whitelist:
   cloudplatform-live1-1: "35.178.209.113/32"
   cloudplatform-live1-2: "3.8.51.207/32"
   cloudplatform-live1-3: "35.177.252.54/32"
+
+community-api:
```
jhevans: I've added these placeholders to prevent deployment failures due to missing properties but a better way might have been to use a feature flag to disable the community api endpoints in environments we don't have credentials for yet. I want to fix the pipeline asap but can go back and  do this if we decide it's worth it

PR #52, opened 2020-03-24
PIC-358: Add crn field to court case

jhevans:


src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerIntTest.java
```diff
@@ -229,7 +229,7 @@ public void createCaseDataWithIncorrectCourt() {
                 .accept(ContentType.JSON)
                 .body(caseDetails)
                 .when()
-                .put("/case/" + CASE_ID)
+                .put("/case/" + "123456")
```
chris-faulkner-mojdt: Why lose the constant ? And there's no point concatenating. IDE mistake ?

jhevans: yep, will amend

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerIntTest.java
```diff
@@ -261,6 +261,7 @@ public void createCaseDataWithDuplicateCaseNo() {
         String newCaseId = "666666";
         caseDetails.setCaseId(newCaseId);
         caseDetails.setCaseNo(CASE_NO);
+        caseDetails.setCrn("CRN");
```
chris-faulkner-mojdt: This line isn't needed in this test. No assertions or dependency on that field being updated. Can remove or use the CRN static final

jhevans: Good spot, I added it in after briefly toying with the idea of making crn non-nullable. Decided not to because we don't actually know what we're going to get from the matcher yet. I raised a story to look into DB hygiene and appropriate constraints once we've got a better view of this though.

PR #54, opened 2020-03-26
360 - empty requirements not working, returning 500

chris-faulkner-mojdt:
Community API returned the list with "requirements" as a property name. 

This depends on an update to community API which is in progress - do not merge yet.


build.gradle
```diff
@@ -77,6 +77,20 @@ test {
         excludeTestsMatching '*SmokeTest'
         excludeTestsMatching '*WIP*'
     }
+
+    testLogging {
```
jhevans: Good call 👍

src/test/java/uk/gov/justice/probation/courtcaseservice/restclient/OffenderRestClientIntTest.java
```diff
@@ -84,4 +86,12 @@ public void whenGetConvictionRequirementsCalled_thenMakeRestCallToCommunityApi()
         assertThat(optionalRequirements.get()).hasSize(3);
     }
 
+    @Test
+    public void whenGetConvictionRequirementsCalled_thenMakeRestCallToCommunityApiEmptyRequirements() {
```
jhevans: This name doesn't really describe what the test is doing, I would suggest the following:
```suggestion
    public void givenEmptyRequirementsWillBeReturnedFromCommunityApi_whenGetConvictionRequirementsCalled_thenReturnEmptyRequirements() {
```

src/test/java/uk/gov/justice/probation/courtcaseservice/restclient/communityapi/mapper/OffenderMapperTest.java
```diff
@@ -128,4 +128,15 @@ public void shouldMapRequirementDetailsToRequirement() {
         assertThat(requirement3.getTerminationDate()).isEqualTo(LocalDate.of(2018,12,01));
         assertThat(requirement3.getRqmntTerminationReasonId()).isEqualTo("2500052884");
     }
+
+    @Test
+    public void shouldMapEmpty() throws IOException {
+        var objectMapper1 = new ObjectMapper();
+        objectMapper1.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        objectMapper1.enable(DeserializationFeature.ACCEPT_EMPTY_ARRAY_AS_NULL_OBJECT);
```
jhevans: Is this right? I'd expect this to do the exact opposite of what we want and turn a valid empty array into a null? Also I can't see this configured globally anywhere so will this actually be applied in the application runtime?

src/test/java/uk/gov/justice/probation/courtcaseservice/restclient/OffenderRestClientIntTest.java
jhevans: This is what I've been going for: https://martinfowler.com/bliki/GivenWhenThen.html

build.gradle
chris-faulkner-mojdt: Thanks. Had been bothering me for a while :-)

src/test/java/uk/gov/justice/probation/courtcaseservice/restclient/communityapi/mapper/OffenderMapperTest.java
chris-faulkner-mojdt: Sorry - that was a hangover from earlier tests. Works without. Actually - in that test class, all 3 files get loaded for all 4 tests, but they're only needed once in each.. Would like to move the file loading into the test methods ? Would be more performant and fewer class fields.

src/test/java/uk/gov/justice/probation/courtcaseservice/restclient/OffenderRestClientIntTest.java
chris-faulkner-mojdt: I copied the existing method and added what happens to the "then" part "whenGetConvictionRequirementsCalled_thenMakeRestCallToCommunityApi" which was already there. I'll change both methods ?

PR #59, opened 2020-03-31
Fix some swagger models and modify requirements endpoint to return ar…

jhevans:
…ray nested in object. Fix createTestData script to accept 201 response as success

Please wait for both reviewers before merging.


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderController.java
```diff
@@ -54,7 +51,7 @@
         })
     @GetMapping(path="offender/{crn}/convictions/{convictionId}/requirements", produces = APPLICATION_JSON_VALUE)
     public @ResponseBody
-    Mono<List<Requirement>> getRequirements(@PathVariable String crn, @PathVariable String convictionId) {
-        return offenderService.getConvictionRequirements(crn, convictionId);
+    RequirementsResponse getRequirements(@PathVariable String crn, @PathVariable String convictionId) {
```
jhevans: @katie-cawthorne I had to change this to an Object rather than returning a list directly as Swagger wasn't able to handle the Mono<List>, hope that's ok. Returning a flat list can cause problems for some clients as well so if possible it's better to just return an object with a single array field.

src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderServiceTest.java
```diff
@@ -38,8 +38,8 @@ public void setUp() {
         service = new OffenderService(offenderRestClient);
         offender = Offender.builder().build();
         when(offenderRestClient.getOffenderByCrn(CRN)).thenReturn(Mono.just(offender));
-        when(offenderRestClient.getConvictionsByCrn(CRN)).thenReturn(Mono.just(convictions));
-        when(offenderRestClient.getConvictionRequirements(CRN, CONVICTION_ID)).thenReturn(Mono.just(requirements));
+        when(offenderRestClient.getConvictionsByCrn(CRN)).thenReturn(Mono.just(expectedConvictions));
```
chris-faulkner-mojdt: Would like to see these mocked control statements moved into just the test methods which use them. ...

PR #63, opened 2020-04-07
PIC-311: Added unpaid work to convictions response

katie-cawthorne:
- Added unpaid work to sentence within the convictions response, in line with community-api
- Updated tests
- Updated mock data


src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/UnpaidWork.java
```diff
@@ -0,0 +1,17 @@
...
+
+import io.swagger.annotations.ApiModel;
+import lombok.Builder;
+import lombok.Getter;
+
+@ApiModel("UnpaidWork")
+@Getter
+@Builder
+public class UnpaidWork {
+    private Integer offered;
```
jhevans, line 11: We should document that this value is given in minutes... in fact maybe just rename it to minutesOffered and not bother with documenting it. Sorry I appreciate this is what it says in the ticket!

src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/communityapi/model/CommunityApiAppointment.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.justice.probation.courtcaseservice.restclient.communityapi.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Getter;
+
+@Getter
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CommunityApiAppointment {
```
jhevans, line 9: Maybe rename this? `CommunityApiAttendance` or something like that might be clearer?  Up to you

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/UnpaidWork.java
katie-cawthorne: I agree, I did think that when I was doing it actually. I'll update it.

src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/communityapi/model/CommunityApiAppointment.java
katie-cawthorne: Sure makes more sense, I'll update it now.

katie-cawthorne: Actually, just had a look and we already have `CommunityApiAttendance` so I'll keep it as `CommunityApiAppointment`, it's actually referred to as `appointment` in community-api.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/UnpaidWork.java
jhevans: Awesome! Don't feel like the ticket is gospel, that was just my best interpretation of what seemed sensible at the time but we should feel free to change it if we find things that can be done better. Just leave a comment or something so people coming after know.

PR #68, opened 2020-04-17
Awaiting Wiremock to be ready. 

chris-faulkner-mojdt:
We looked to be hitting the exact same issue described here 
[https://github.com/tomakehurst/wiremock/issues/1224](url)

This ticket reports this 

> we sometimes have failures once Spring launches a fresh context and Spring Security tries to get OAuth2 token (this call is stubbed within WireMock server)

The workaround was to do this

> We now have a workaround and add a simple e.g. /readiness/wiremock stub right after starting the WireMockServer and try to hit it with a simple retry-mechanism and once HTTP status 200 returns we continue we our setup.

The ticket was closed with no fix. So this commit does the same thing as described

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -172,6 +172,7 @@ dependencies {
     }
 
     testCompile "com.github.tomakehurst:wiremock-jre8:2.26.2"
+    testCompile("org.springframework.retry:spring-retry:1.2.5.RELEASE")
 
     implementation ('org.springframework:spring-web:5.2.3.RELEASE') {
         because 'Transitive dependency. Previous versions have a known vulnerability'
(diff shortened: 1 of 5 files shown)


jhevans:
> We looked to be hitting the exact same issue described here
> [https://github.com/tomakehurst/wiremock/issues/1224](url)
> 
> This ticket reports this
> 
> > we sometimes have failures once Spring launches a fresh context and Spring Security tries to get OAuth2 token (this call is stubbed within WireMock server)
> 
> The workaround was to do this
> 
> > We now have a workaround and add a simple e.g. /readiness/wiremock stub right after starting the WireMockServer and try to hit it with a simple retry-mechanism and once HTTP status 200 returns we continue we our setup.
> 
> The ticket was closed with no fix. So this commit does the same thing as described

I get a 404 when I follow that link

PR #70, opened 2020-04-21
PIC-401 add assessment details to 'probation-record' endpoint

tomsmyers:
the only real logic in these changes is to do with gracefully handing errors from the assessments api (or client) so that we can still populate the report even without the assessment data.

one other change worth noting is that to get the `offender-assessments-api` service running locally with auth working requires a one time change to your `etc/hosts` file. this is detailed in the readme.

coverage on new code is at 100%. the only place we are not able to get coverage are lines 35-36 of `RestClientHelper.java`, which applies the auth token to the outgoing http requests. side note on coverage: why aren't we seeing coverage reports in sonar? i checked all the jacoco and sonar config and it looks right.


src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderService.java
```diff
@@ -22,23 +24,31 @@
...
     }
 
     public ProbationRecord getProbationRecord(String crn, boolean applyDocumentFilter) {
-
-        Tuple3<ProbationRecord, List<Conviction>, GroupedDocuments> tuple3 = Mono.zip(client.getProbationRecordByCrn(crn), client.getConvictionsByCrn(crn), client.getDocumentsByCrn(crn))
+        Tuple4<ProbationRecord, List<Conviction>, GroupedDocuments, Assessment> tuple4 = Mono.zip(
+            defaultClient.getProbationRecordByCrn(crn),
+            defaultClient.getConvictionsByCrn(crn),
+            defaultClient.getDocumentsByCrn(crn),
+            assessmentsClient.getAssessmentByCrn(crn))
```
tomsmyers, line 45: should we handle errors from the assessments api separately here?

if the CRN isn't found in oasys should we still return the rest of the convictions data without the assessment attached?

chris-faulkner-mojdt:
In terms of test coverage - there's a branch to fix which I haven't had chance to return to.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderService.java
```diff
@@ -22,24 +24,42 @@
...
-            .blockOptional()
-            .orElseThrow(() -> new OffenderNotFoundException(crn));
-
+        var tuple3 = probationMono.blockOptional().orElseThrow(() -> new OffenderNotFoundException(crn));
         ProbationRecord probationRecord = addConvictionsToProbationRecord(tuple3.getT1(), tuple3.getT2());
         combineConvictionsAndDocuments(probationRecord, tuple3.getT3().getConvictions(), applyDocumentFilter);
 
+        try {
+            assessmentMono.blockOptional().ifPresent(assessment -> probationRecord.setAssessment(assessment));
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
jhevans: Never mind, have seen it throws checked exceptions in the service. Probably better to be specific though. I appreciate we want it to continue if the assessment isn't found but we don't want it swallowing NullPointerExceptions and the like that we genuinely don't expect.

src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderServiceTest.java
```diff
@@ -145,4 +159,46 @@ public void whenGetConvictionRequirements_returnRequirements() {
...
+        // the assessment field should just be empty
+        assertThat(probationRecord.getAssessment()).isNull();
+
+        // but the rest of the record should be populated as normal
+        assertThat(probationRecord.getConvictions()).hasSize(1);
+        final Conviction conviction = probationRecord.getConvictions().get(0);
+        assertThat(conviction.getDocuments()).hasSize(2);
+    }
+
+    @DisplayName("Getting probation record does not throw exception when assessment api fails for any reason")
```
jhevans, line 184: Hmm, not sure about this. The danger is we a) suppress errors that we need to fix and b) give the user the impression that data doesn't exist in the system when in practice it just isn't available due to an error. 

The 'a' we can mitigate by logging an error I think 'b' is more dangerous, we don't want to give probation staff misleading information as that might ultimately affect what sentence is passed.

src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/RestClientHelper.java
```diff
@@ -36,6 +36,9 @@
         return spec.attributes(clientRegistrationId(oauthClient));
     }
 
+    // handleError is a slightly modified version of the default error handler which returns
+    // `OffernderNotFoundException` for 404 status codes, and `WebClientResponseException` for
```
jhevans, line 40:
```suggestion
    // `OffenderNotFoundException` for 404 status codes, and `WebClientResponseException` for
```

chris-faulkner-mojdt:
OK

PR #78, opened 2020-05-06
PIC-380 add breaches to convictions in probation record response

tomsmyers:


src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/communityapi/mapper/interventions/BreachMapper.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.stream.Collectors;
+
+import org.springframework.stereotype.Component;
+import uk.gov.justice.probation.courtcaseservice.service.model.Breach;
+import uk.gov.justice.probation.courtcaseservice.restclient.communityapi.model.CommunityApiNsi;
+import uk.gov.justice.probation.courtcaseservice.restclient.communityapi.model.CommunityApiNsiResponse;
+
+@Component
+public class BreachMapper {
+    public Breach breachFrom(CommunityApiNsi nsi) {
```
chris-faulkner-mojdt, line 13: Would expect a unit test class direct on this. It's probably effectively tested by integration tests higher up additionally but please add

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/Breach.java
```diff
@@ -0,0 +1,20 @@
...
+import java.time.LocalDate;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@ApiModel("Breach")
+@Getter
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Breach {
+    private String id;
```
chris-faulkner-mojdt, line 16: Why string here ? The community API has the nsi id as a long ? Also, lower down, the treatment of convictionId as a string is an overhead in the comparable implementation (although not part of this change)

src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderServiceTest.java
```diff
@@ -83,6 +92,7 @@ void beforeEach() {
     void whenGetOffender_returnOffenderWithConvictionsDocumentsNotFiltered() {
         when(offenderRestClient.getProbationRecordByCrn(CRN)).thenReturn(Mono.just(ProbationRecord.builder().crn(CRN).build()));
         when(offenderRestClient.getConvictionsByCrn(CRN)).thenReturn(Mono.just(singletonList(conviction)));
+        when(offenderRestClient.getBreaches(CRN, CONVICTION_ID)).thenReturn(Mono.just(singletonList(breach)));
```
chris-faulkner-mojdt, line 95: Would like to see the Mockito verify call to prove that this happened. Mockito runs in non-strict mode and it's good to know that each and every "when" is necessary and doing as expected.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/Breach.java
tomsmyers: i chose String because that seems to be the representation for ids in other models e.g. `documentId` `convictionId`.

agree there is some overhead in converting convictionId for comparisons, but i didn't want to change the existing conviction model.

src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderServiceTest.java
tomsmyers: can do, although mockito does enforce this anyway. if you add a stub for something that isn't called the test fails with an exception like this:

```
org.mockito.exceptions.misusing.UnnecessaryStubbingException: 
Unnecessary stubbings detected.
Clean & maintainable test code requires zero unnecessary code.
Following stubbings are unnecessary (click to navigate to relevant line of code):
  1. -> at uk.gov.justice.probation.courtcaseservice.service.OffenderServiceTest.givenOffenderNotFound_whenGetOffender_thenThrowException(OffenderServiceTest.java:144)
```

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/Breach.java
chris-faulkner-mojdt: OK - understand the reason not to change the conviction. documentId really is a string - it's a GUID type thing. I'd like to see us using a long unless there's a reason for string

tomsmyers: done

PR #79, opened 2020-05-14
Pic 387 add get breach endpoint

jhevans:
Implementation of [PIC-387](https://dsdmoj.atlassian.net/browse/PIC-387)

Additionally I've done some refactoring to introduce a common RestResourceNotFound class which the existing OffenderResourceNotFound class now extends. This is handled generically in the application exception handler but allows us to be specific about the resource we're missing. I've used this to return a ConvictionNotFound exception in a couple of places where we we're erroneously throwing for Offender - I think this is better but would welcome comments on this one in particular for cases like 'convictions/attendances' where it's possibly not quite clear cut. We could introduce more exceptions for 'Attendances not found' but I figured that would actually be an empty list so conviction not found was right.

Just a note on something that I noticed that I think we probably need to address. We've got some inconsistency with our endpoints - for example some start `offender` and others `offenders`. Though we fairly consistently refer to 'convictions'. We could do with aligning these at some point but for now I've gone with 'offender' for the new endpoint as it's the most commonly used.

Another interesting point to note - because it's using Mono.zip to make the two calls for the conviction and nsi, either could return first. This means you could get two possible errors (Nsi not found, or Conviction not found). I'm not sure how happy I am about that, comments welcome. Should we prefer determinism or performance in the happy case?


src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/NsiRestClient.java
```diff
@@ -0,0 +1,26 @@
...
+    @Autowired
+    @Qualifier("communityApiClient")
+    private RestClientHelper clientHelper;
+    @Value("${community-api.nsi-by-id-url-template}")
+    private String nsiUrlTemplate;
+
+    public Mono<CommunityApiNsi> getNsiById(String crn, Long convictionId, Long nsiId) {
+        return clientHelper.get(String.format(nsiUrlTemplate, crn, convictionId, nsiId))
+                .retrieve()
+                .onStatus(HttpStatus::is4xxClientError, (clientResponse) -> clientHelper.handleNsiError(crn, convictionId, nsiId, clientResponse))
```
chris-faulkner-mojdt, line 23: There are 2 integration tests for NsiRestClient. In the other *ClientIntTests, we tend to have at least 3 with the 3rd ensuring that the WebClientException is expected from one (for the 500s).  
`@Test(expected = WebClientResponseException.class)`

chris-faulkner-mojdt:
Good stuff with the new RestRourceNotFoundException and the other refactorings.

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderControllerIntTest.java
```diff
@@ -180,6 +175,36 @@ public void whenCallMadeToGetRequirementData_thenReturnCorrectData() {
 
     }
 
+    @Test
```
chris-faulkner-mojdt, line 178: No test for server error here

src/test/java/uk/gov/justice/probation/courtcaseservice/service/BreachServiceTest.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Test
+    public void whenGetBreachExists_thenReturnBreach() {
+        when(nsiRestClient.getNsiById(CRN, CONVICTION_ID, BREACH_ID)).thenReturn(Mono.just(nsi));
+        when(convictionRestClient.getConviction(CRN, CONVICTION_ID)).thenReturn(Mono.just(conviction));
+        when(nsiMapper.breachOf(nsi, conviction)).thenReturn(breachResponse);
+        BreachResponse breachResponse = breachService.getBreach(CRN, CONVICTION_ID, BREACH_ID);
+
+        assertThat(breachResponse).isEqualTo(breachResponse);
```
chris-faulkner-mojdt, line 51: Tiny readability one - breachResponse is hiding the class level one. this.breachResponse on second ?

src/test/resources/http-request/breach-ctrl.rest
```diff
@@ -0,0 +1,55 @@
+### Known breach for a known conviction and offender
```
chris-faulkner-mojdt, line 1: Much, much appreciated you doing this, and I know you can't even run them. Thanks.

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/BreachResponse.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.justice.probation.courtcaseservice.controller.model;
+
+import lombok.Builder;
+import lombok.Getter;
+
+import java.time.LocalDate;
+
+@Builder
+@Getter
+public class BreachResponse {
```
chris-faulkner-mojdt, line 10: Need that non-null annotation, can get responses like this 

`{
  "breachId": 2500018597,
  "incidentDate": "2019-09-02",
  "started": "2019-10-02",
  "provider": "NPS North East",
  "team": "Unallocated Team(N02)",
  "officer": "Unallocated Staff(N02) Staff",
  "status": null,
  "order": "CJA - Indeterminate Public Prot."
}`

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderControllerTest.java
```diff
@@ -1,64 +1,106 @@
...
     @Mock
     private Requirement expectedRequirement;
+    @Mock
+    private ConvictionService convictionService;
+    @Mock
+    private BreachService breachService;
+
+    private FeatureFlags featureFlags;
 
-    @InjectMocks
```
chris-faulkner-mojdt, line 33: Why lose this ? Eliminates the need for set up.

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderControllerTest.java
```diff
@@ -1,64 +1,106 @@
...
-    private OffenderService service;
+    private OffenderService offenderService;
     @Mock
     private ProbationRecord expectedProbationRecord;
     @Mock
     private Requirement expectedRequirement;
+    @Mock
+    private ConvictionService convictionService;
+    @Mock
+    private BreachService breachService;
```
chris-faulkner-mojdt, line 39: Missing test. This field is only used in the constructor for OffenderController. When I run OffenderControllerTest with coverage, there are no tests over getBreach.

src/test/resources/http-request/breach-ctrl.rest
jhevans: I can for now.... I've been running on an evaluation license for what must be well over 30 days now 🤷 I've been told there is a batch of licenses on it's way though so 🤞

src/test/java/uk/gov/justice/probation/courtcaseservice/service/BreachServiceTest.java
jhevans: The assertion is actually comparing the same variable as well 🤦 have updated to expected/actual and fixed. Good spot!

src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/NsiRestClient.java
chris-faulkner-mojdt: Like the alternative way of expressing

PR #90, opened 2020-06-03
:sparkles:  518 PUT by court and case no

chris-faulkner-mojdt:
The PUT to create / update by case ID is deprecated. We now update by what we are thinking of as our unique business key (court code and case no). The headline of this change is a new method that does pretty much the same as the PUT by case ID. Other associated changes

- make a unique key of court code / case no. These fields should be immutable once saved so no setters on the Entity object
- case no is no longer unique by itself

### DIFF
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/service/CourtCaseService.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/service/CourtCaseService.java
@@ -1,5 +1,6 @@
 package uk.gov.justice.probation.courtcaseservice.service;
 
+import java.util.Optional;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/justice/probation/courtcaseservice/service/CourtCaseService.java
```diff
@@ -73,21 +97,15 @@ public CourtCaseEntity createOrUpdateCase(String caseId, CourtCaseEntity courtCa
         existingCase.setBreach(courtCaseEntity.getBreach());
         existingCase.setDefendantName(courtCaseEntity.getDefendantName());
         existingCase.setDefendantAddress(courtCaseEntity.getDefendantAddress());
-
-        log.info("Court case updated for case {}", courtCaseEntity.getCaseNo());
+        existingCase.setPnc(courtCaseEntity.getPnc());
```
chris-faulkner-mojdt, line 100: I aded these fields to be set because they weren't orginally. No tests existed to send updates based on those fields.

PR #96, opened 2020-06-17
:bug: restore initial v108 behaviour

chris-faulkner-mojdt:


src/main/resources/db/migration/courtcase/V108__make_unique_key_for_court_and_case_no_court_case.sql
```diff
@@ -1,5 +1,5 @@
 ALTER TABLE COURT_CASE DROP CONSTRAINT court_case_case_no_idempotent;
 
-CREATE UNIQUE INDEX court_case_court_case_uq ON COURT_CASE (CASE_NO, COURT_CODE);
+CREATE UNIQUE INDEX court_case_court_case_uq ON COURT_CASE (COURT_CODE, CASE_NO);
```
jhevans, line 3: I assume the ordering here doesn't actually matter and that we don't need to apply this change in a later migration?

chris-faulkner-mojdt: If you put the most selective item first it improves performance. It's not that important given our likely data volumes.

PR #97, opened 2020-06-18
:bug: make case_no nullable to start

chris-faulkner-mojdt:


src/main/resources/db/migration/courtcase/V109__make_fk_from_offence_to_court_case.sql
```diff
@@ -5,8 +5,14 @@ ALTER TABLE OFFENCE DROP COLUMN case_id;
 ALTER TABLE COURT_CASE DROP CONSTRAINT court_case_case_id_idempotent;
 
 -- Add columns to make the FK
-ALTER TABLE OFFENCE ADD COLUMN CASE_NO TEXT NOT NULL;
-ALTER TABLE OFFENCE ADD COLUMN COURT_CODE TEXT NOT NULL;
+ALTER TABLE OFFENCE ADD COLUMN CASE_NO TEXT;
+ALTER TABLE OFFENCE ADD COLUMN COURT_CODE TEXT;
+
+DELETE FROM OFFENCE WHERE ID > 1;
```
jhevans, line 11: On this line I think you could alternatively use `TRUNCATE TABLE OFFENCE;` rather than using a conditional delete but it does the same job. I would expect that truncate is probably better optimised if dealing with large amounts of data but we've probably only got a few dozen records so thats a non-issue here.

src/main/resources/db/migration/courtcase/V109__make_fk_from_offence_to_court_case.sql
```diff
@@ -5,8 +5,14 @@ ALTER TABLE OFFENCE DROP COLUMN case_id;
 ALTER TABLE COURT_CASE DROP CONSTRAINT court_case_case_id_idempotent;
 
 -- Add columns to make the FK
-ALTER TABLE OFFENCE ADD COLUMN CASE_NO TEXT NOT NULL;
-ALTER TABLE OFFENCE ADD COLUMN COURT_CODE TEXT NOT NULL;
+ALTER TABLE OFFENCE ADD COLUMN CASE_NO TEXT;
```
jhevans, line 8: Correct me if I'm wrong but I don't think you'd need these steps in this instance because by the time you come to add the `NOT NULL` constraint at line 14 the table is already empty because of the delete on line 11?

src/main/resources/db/migration/courtcase/V109__make_fk_from_offence_to_court_case.sql
chris-faulkner-mojdt: That's right . I only use TRUNCATE in exceptional circumstances and sometimes delete doesn't work without a where clause

src/main/resources/db/migration/courtcase/V109__make_fk_from_offence_to_court_case.sql
chris-faulkner-mojdt: That's right. :-) What can I say - still early ?

PR #107, opened 2020-07-09
:sparkles: Add soft deleted and audit fields

chris-faulkner-mojdt:
Created a BaseEntity. We could consider moving the ID field in there. Needed SuperBuilder instead of Builder in Lombok annotation. Added a delete method in CourtCaseService and an integration test for it.


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/BaseEntity.java
```diff
@@ -0,0 +1,48 @@
...
+    @LastModifiedDate
+    private LocalDateTime lastUpdated;
+
+    @Column(name = "created_by", nullable = true, updatable = false)
+    @CreatedBy
+    private String createdBy;
+
+    @Column(name = "last_updated_by")
+    @LastModifiedBy
+    private String lastUpdatedBy;
```
jhevans, line 40: Out of curiosity, does this work automagically with Spring security by getting it from the Principal or will we have to implement this?

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CourtCaseEntity.java
```diff
@@ -75,7 +76,7 @@
 
     @JsonManagedReference
     @OneToMany(mappedBy = "courtCase", fetch = FetchType.EAGER, cascade = { CascadeType.ALL }, orphanRemoval=true)
-    private List<OffenceEntity> offences = new ArrayList<>();
+    private List<OffenceEntity> offences;
```
jhevans, line 79: 👍  I assume this is to prevent that compiler warning, good work - I've been meaning to do the same but didn't want to risk breaking stuff 😄

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/BaseEntity.java
chris-faulkner-mojdt: Yes - if / when we have spring security then it will / should "just work"

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CourtCaseEntity.java
chris-faulkner-mojdt: Yes

PR #115, opened 2020-07-18
Pic 526 Auth

jhevans:


jhevans:
I removed the FeatureFlags endpoint because we're not currently using it and it didn't seem appropriate to grant that permission to `ROLE_PREPARE_A_CASE` which is the only role we currently have available as it's an internal administrative thing not an end user feature. I also didn't want to leave it unauthenticated because it just increases our attack surface. I'd suggest we don't squash the commits for this PR if possible so that we can simply revert the commit to remove FeatureFlags if we decide to reimplement in future.

PR #117, opened 2020-07-21
✨ Allow missing cases for a date to be deleted

chris-faulkner-mojdt:
The code takes a list of case numbers for a date / court code and deletes any cases found for that court / date which are not n the list of case numbers. The list of case numbers has arrived and is the full list of cases for that date / court.

The approach is to use SQL to get a list of cases to delete and then use a standard Spring-provided repository delete method to delete them. Due to work done earlier, this is a "soft delete" and cascades to child offences. Other approaches were looked at 

- Custom query to to a DELETE. Unfortunately it's a feature of JPA that this doesn't cascade to child entities and doesn't understand the soft delete. SO we'd have had to write it as an UPDATE on both the COURT_CASE and OFFENCE tables.
- CriteriaBuilder - again this would have to be written as an UPDATE. I trialled this approach and whilst it works, it results in having to manage transactions with the EntityManager and very many more lines of code.  The CriteriaBuilder tools are useful for other scenarios and might be better if we were looking at large batch deletes, at least dozens of records. We are probably looking at a maximum of one or two deletes per transaction and in many cases, zero.

### DIFF
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
@@ -97,7 +97,6 @@ public class CourtCaseController {
     @ApiOperation(value = "For the date / cases passed, delete any cases NOT in the list.")
     @ApiResponses(
         value = {
-            @ApiResponse(code = 204, message = "OK", response = CaseListResponse.class),
             @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
             @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class),
             @ApiResponse(code = 403, message = "Forbidden", response = ErrorResponse.class),
@@ -105,8 +104,8 @@ public class CourtCaseController {
             @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
         })
     @ResponseStatus(value = HttpStatus.NO_CONTENT)
-    @PutMapping(value = "/court/{courtCode}/cases/purge", consumes = APPLICATION_JSON_VALUE, produces = APPLICATION_JSON_VALUE)
-    public void deleteCasesByDate(@PathVariable String courtCode, @Valid @RequestBody Map<LocalDate, List<String>> existingCasesByDate) {
-        courtCaseService.deleteMissingCases(courtCode, existingCasesByDate);
+    @PutMapping(value = "/court/{courtCode}/cases/purgeAbsent", consumes = APPLICATION_JSON_VALUE, produces = APPLICATION_JSON_VALUE)
+    public void purgeAbsentCases(@PathVariable String courtCode, @Valid @RequestBody Map<LocalDate, List<String>> existingCasesByDate) {
+        courtCaseService.deleteAbsentCases(courtCode, existingCasesByDate);
     }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
```diff
@@ -93,4 +93,20 @@ CaseListResponse getCaseList(@PathVariable String courtCode,
                 .collect(Collectors.toList());
         return CaseListResponse.builder().cases(courtCaseResponses).build();
     }
+
+    @ApiOperation(value = "For the date / cases passed, delete any cases NOT in the list.")
+    @ApiResponses(
+        value = {
+            @ApiResponse(code = 204, message = "OK", response = CaseListResponse.class),
```
jhevans, line 100: Response class is incorrect, think we can remove this line as it's derived automatically anyway and very easy to get wrong!

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerPutIntTest.java
```diff
@@ -67,6 +77,11 @@
     private static final String JSON_CASE_NO = "1700028914";
     private static final String JSON_CASE_ID = "654321";
 
+//    @ClassRule
```
jhevans, line 80: 🔥 Commented code can be removed

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
```diff
@@ -93,4 +93,20 @@ CaseListResponse getCaseList(@PathVariable String courtCode,
...
+            @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+            @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class),
+            @ApiResponse(code = 403, message = "Forbidden", response = ErrorResponse.class),
+            @ApiResponse(code = 404, message = "If the court is not found by the code passed."),
+            @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
+        })
+    @ResponseStatus(value = HttpStatus.NO_CONTENT)
+    @PutMapping(value = "/court/{courtCode}/cases/purge", consumes = APPLICATION_JSON_VALUE, produces = APPLICATION_JSON_VALUE)
```
jhevans, line 108: Just wondering whether `purgeAbsent/missing` would be more a more descriptive name? Honestly not sure as it may raise more questions than it answers. The swagger description is very clear which is the main thing - up to you if you think this makes it clearer or not.

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
chris-faulkner-mojdt: Done

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerPutIntTest.java
chris-faulkner-mojdt: Done

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
chris-faulkner-mojdt: It's a good point. Yes - earlier I was thinking along those lines because I named the service method "deleteMissingCases". I'll standardise around your "absent". Kept deleteAbsentCases in service and repo because the "delete" matches what happens in repo.

PR #125, opened 2020-07-29
Pic 629 health check

jhevans:


src/main/java/uk/gov/justice/probation/courtcaseservice/application/WebClientConfig.java
```diff
@@ -24,26 +23,33 @@
...
     public WebClient assessmentsWebClient(OAuth2AuthorizedClientManager authorizedClientManager) {
         return webClientFactory(assessmentsApiBaseUrl, authorizedClientManager);
     }
 
+    @Bean
+    public WebClient oauthWebClient(OAuth2AuthorizedClientManager authorizedClientManager) {
+        return webClientFactory(oauthApiBaseUrl, authorizedClientManager);
+    }
+
     private WebClient webClientFactory(String baseUrl, OAuth2AuthorizedClientManager authorizedClientManager) {
```
chris-faulkner-mojdt, line 53: in CCM, we overwrote the web client with configurable timeouts - see code snippet below. I think we should adopt this now. The delays we have seen in preprod are much too long for a user.

I think that we should have shorter timeouts for our "pinging" clients. Especially for the read timeout.

`        HttpClient httpClient = HttpClient.create()
            .tcpConfiguration(client ->
                client.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectTimeoutMs)
                    .doOnConnected(conn -> conn
                        .addHandlerLast(new ReadTimeoutHandler(readTimeoutMs, TimeUnit.MILLISECONDS))
                        .addHandlerLast(new WriteTimeoutHandler(writeTimeoutMs, TimeUnit.MILLISECONDS))));`

PR #136, opened 2020-08-19
👷‍♀️ Prevent build failures for dependabot builds

jhevans:


theDustRoom (approved):
Hmm. Docker isn't building still

jhevans:
I've gone with the short hash rather than branch name as there's a hard limit of 128 characters for tag length, and with these dependabot PRs it's only a matter of time before we hit that I reckon. Short hash is traceable to an exact commit, the only disadvantage is we lose human readability - in my experience we very rarely have anything but the trunk version deployed anyway so this doesn't seem like too much of a problem but let me know if you think otherwise. @katie-cawthorne @theDustRoom @chris-faulkner-mojdt FYI

chris-faulkner-mojdt (approved):
LGTM

PR #146, opened 2020-08-26
🔒 Disable health and info endpoints in prod

jhevans:


helm_deploy/court-case-service/templates/_envs.tpl
```diff
@@ -8,7 +8,7 @@ env:
     value: "{{ .Values.image.port }}"
 
   - name: SPRING_PROFILES_ACTIVE
-    value: "postgres,instrumented"
```
jhevans, line 11: Note: the postgres profile was removed ages back because it didn't actually do anything so I've removed it from the  values files accordingly

chris-faulkner-mojdt (approved):
LGTM

PR #169, opened 2020-08-28
🐛 Fix unhandled error case where convictions call returns not found w…

jhevans:
…hen getting offenderMatchDetail

### DIFF
--- a/src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderMatchServiceTest.java
+++ b/src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderMatchServiceTest.java
@@ -1,12 +1,6 @@
 package uk.gov.justice.probation.courtcaseservice.service;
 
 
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderMatchServiceTest.java
```diff
@@ -1,12 +1,6 @@
 package uk.gov.justice.probation.courtcaseservice.service;
 
 
-import java.time.LocalDate;
```
chris-faulkner-mojdt, line 4: Not massively important. I changed my import order so that java.* and javax.* would be at the top. I thought this would match yours - I don't really mind which it is but it would be good if we synchronised on this. Just to reduce noise in the PRs

chris-faulkner-mojdt (approved):
LGTM

PR #170, opened 2020-08-28
🔧 Update ingress config to match cloud platform updates

jhevans:
This fix is in line with this advice from Cloud Platform regarding the ongoing ingress changes: https://mojdt.slack.com/archives/CH6D099DF/p1598458623008500

By applying these changes the existing endpoints should continue to work with no further changes


chris-faulkner-mojdt (approved):
LGTM

PR #171, opened 2020-08-28
 🔧 Update ingress controller config in dev and preprod to use namespa…

jhevans:
…ce local controller


chris-faulkner-mojdt (approved):
LGTM

PR #180, opened 2020-09-03
Readme update

sarahyjja:
As I had quite some difficulties to have this repo up and running on my machine, I would like to propose you a new readme file which might be easier for the next newbie in the project. Please feel free to correct anything which doesn't look right :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,99 +1,121 @@
-Court Case Service
-==================
-Service to access court cases imported from HMCTS Libra court lists
+# Court Case Service
+### Service to access court cases imported from HMCTS Libra court lists
 
-Dev Setup
 ---
 
-In order to run the service locally, a postgres database and some other backend services are required, the easiest way to run locally is using the [docker-compose.yml](docker-compose.yml) file which will pull down the latest version.
+## Setup your machine
 
-```docker-compose up```
+In the root folder of your machine start by checking if HomeBrew is installed on your computer with the command : `brew -version` 
+If you have HomeBrew and Java set up already, go to the Dependencies section. 
+:::info 
+⚠️ This project run with ==JAVA 11== and NOT the last version!
+:::
 
-In order for the authentication to function between services locally you must add the following line to your `/etc/hosts` file before starting:
 
-```
-127.0.0.1 oauth
-```
 
-The service uses Lombok and so annotation processors must be [turned on within the IDE](https://www.baeldung.com/lombok-ide).
+#### 1. If after running the brew version nothing appears do the following :
+~1.1.~ `$ ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
+~1.2.~ `$ brew install`
+~1.3.~ `$ brew tap homebrew/cask`
+~1.4.~ `$ brew tap homebrew/cask-versions`
 
-Building and running
----
+#### 2. Check if the version 11 of Java is here :
+~2.1.~ `$ brew cask info java`
 
-This service is built using Gradle. In order to build the project from the command line, run
+If nothing appears, run :
+~2.2.~ `$ brew cask install java11`
 
-```./gradlew build```
+#### 3. If you have others/multiple versions of Java and you need to switch in between/manage them, then you have to install another tool called jEnv : 
+~3.1.~ `$ brew install jenv`
 
-To run the service, ensure there is an instance of Postgres running and then run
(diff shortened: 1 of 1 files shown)


sarahyjja:
Please, don't pay attention to all the commits I've done previously, I had a hard time to connect to my github account :/

README.md
```diff
@@ -1,99 +1,122 @@
...
-Dev Setup
 ---
 
-In order to run the service locally, a postgres database and some other backend services are required, the easiest way to run locally is using the [docker-compose.yml](docker-compose.yml) file which will pull down the latest version.
 
-```docker-compose up```
+## Setup your machine
 
-In order for the authentication to function between services locally you must add the following line to your `/etc/hosts` file before starting:
+In the root folder of your machine start by checking if HomeBrew is installed on your computer with the command : `brew -version`
```
jhevans, line 10: The bit about Java 11 is definitely useful but the rest might be a bit more detail than we'd normally go into. It's useful stuff but different users might have different ways of installing dependencies (for example Windows/Linux users) and might encounter different problems so trying to document for every possibility quickly becomes intractable!  

Happy to leave it in as it as we don't have huge amounts of documentation, you're right we've been a bit lax on that so I'm very glad you're taking the time to make improvements!

README.md
```diff
@@ -1,99 +1,122 @@
...
+**Application Ping** : `$ curl -X GET http://localhost:8080/ping`
+**Application Feature Flags** : `$ curl -X GET http://localhost:8080/feature-flags`
+**To Check dependency versions** : `$ ./gradlew dependencyUpdates`
 
-### Swagger UI
-The project builds swagger specifications which can be examined and tested via the Swagger UI.
+#### *Flyway commands*
+**Migrate database** : `$ gradle flywayMigrate -i`
+**View details and status information about all migrations** : `$ gradle flywayInfo`
+**Baseline an existing database, excluding all migrations up to and including baselineVersion** :  `$ gradle flywayBaseline`
```
jhevans, line 71: Because Spring Boot automatically bootstraps our database we don't generally use the Gradle commands much (though it's still useful to have) - the one exception to this is `flywayClean` which cleans away all migrations applied to a database and removes the flyway schema. This can be useful if you somehow manage to corrupt your migrations, which I've managed a fair few times 😄 . 

Worth noting that we usually use the provided `./gradlew` (Gradle wrapper) file rather than `gradle`. This ensures you've always got the right version of gradle for a given repo but is otherwise identical in usage.

```suggestion
**Baseline an existing database, excluding all migrations up to and including baselineVersion** :  `$ ./gradlew flywayBaseline`
**Clean schema** : `$ ./gradlew flywayClean`
```

README.md
```diff
@@ -1,99 +1,122 @@
...
+**Application health** : `$ curl -X GET http://localhost:8080/health`
+**Application info** : `$ curl -X GET http://localhost:8080/info`
+**Application Ping** : `$ curl -X GET http://localhost:8080/ping`
+**Application Feature Flags** : `$ curl -X GET http://localhost:8080/feature-flags`
+**To Check dependency versions** : `$ ./gradlew dependencyUpdates`
 
-### Swagger UI
-The project builds swagger specifications which can be examined and tested via the Swagger UI.
+#### *Flyway commands*
+**Migrate database** : `$ gradle flywayMigrate -i`
```
jhevans, line 69:
```suggestion
**Migrate database** : `$ ./gradlew flywayMigrate -i`
```

README.md
```diff
@@ -1,99 +1,122 @@
...
+**View details and status information about all migrations** : `$ gradle flywayInfo`
```
jhevans, line 70:
```suggestion
**View details and status information about all migrations** : `$ ./gradlew flywayInfo`
```

jhevans (changes requested):
Couple of comments and minor in line amendments but happy to approve once they've been reviewed

README.md
sarahyjja: How about making a section for Mac and leave an open door for people Window/Linux? Maybe the next after me will be happy to fill these section :) ?

jhevans (approved):

README.md
jhevans: I'd leave it as it is, we only use macs anyway so it was mostly to illustrate a point, if some intrepid Windows user does come along and then they can of course submit a PR!

jhevans (approved):

PR #199, opened 2020-09-18
✨ Added PSS requirements to the existing requirement endpoint

chris-faulkner-mojdt:
As agreed PSS requirements is added to the existing endpoint.
Additional to the original ticket is that subtype description is also returned, but only where the maindescription is "Specified Activity". I have added that to the ticket.


jhevans (approved):
👍

PR #202, opened 2020-09-21
🐛 Fixed handling of a null subtype for PSS requirements

chris-faulkner-mojdt:
Fixed null pointer in subtype mapping of PSS


jhevans (approved):
👍

PR #203, opened 2020-09-21
🐛 Fix sentence endDate calculation

jhevans:


src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderControllerIntTest.java
```diff
@@ -102,6 +102,8 @@ public void whenCallMadeToGetProbationRecord_thenReturnCorrectData() {
                 .body("convictions[2].documents", hasSize(1))
                 .body("convictions[2].documents[0].documentId", equalTo("1d842fce-ec2d-45dc-ac9a-748d3076ca6b"))
                 .body("convictions[2].breaches", hasSize(0))
+                .body("convictions[2].endDate", equalTo(standardDateOf(2019, 1,1)))
```
jhevans, line 105: I've duplicated this logic so that we don't break the front end by removing a field it's expecting. Have raised https://dsdmoj.atlassian.net/browse/PIC-758 to remove it once prepare a case is updated.

katie-cawthorne (approved):
LGTM

PR #211, opened 2020-09-29
✏️ fixed case of "Previously known"

chris-faulkner-mojdt:


katie-cawthorne (dismissed):
LGTM

jhevans (changes requested):
Could you add a data migration to update all existing records to the new value? Would be preferable to doing it manually

chris-faulkner-mojdt:
> Could you add a data migration to update all existing records to the new value? Would be preferable to doing it manually

update courtcaseservice.court_case set probation_status  = 'Previously known' where probation_status = 'Previously Known';

I have done it in prod already. Turned off auto commit first :-)

jhevans (dismissed):

jhevans (approved):

PR #222, opened 2020-10-08
✨ PIC-803 added name and defendant type

chris-faulkner-mojdt:
defendant type is pretty straightforward. It's an enum 

"defendantName" is kept as the existing name field. 
"name" is the new composite. It is saved but not published (yet) in the CourtCaseResponse

Also sorted out the R_Seed_data for new databases


jhevans (approved):
LGTM 👍

PR #227, opened 2020-10-12
👷‍♂️ PIC-826: Enable CircleCI Jira integration

jhevans:
See [ticket](https://dsdmoj.atlassian.net/browse/PIC-826) for proof it's working.

Note that it isn't clever enough to distinguish between different jobs in a build - as far as Jira is concerned it's either built or it isn't. Because of this I'm reporting only on build and docker-build which are sequential so we won't get a situation where a parallel job succeeds and overrides a previously reported and valid failure status. If the app scan fails we won't see it in Jira but this seems to just be a limitation of the integration.


chris-faulkner-mojdt (approved):
LGTM

PR #231, opened 2020-10-14
Pic 570 update docker image and java version

jhevans:
Upgrade to Java 14 (15 isn't supported by Gradle  yet). Also updated the Postgres image to 11 to match what we're deploying to in cloud platform. 

I've changed the Readme a bit because it assumes we run against dockerised dependencies rather than dev which doesn't actually reflect how we generally work with it. For my own part the reason I don't is because I don't have a token to authenticate with the dockerised oauth which is a major limiting factor. There was also a lot of stuff about specific machine set up which I wasn't in a position to re-do and re-test so I stripped it back to the bare minimum of what's needed and put the onus on the reader to sort those dependencies.


docker-compose.yml
```diff
@@ -15,48 +15,48 @@ services:
...
+#    container_name: oauth
+#    ports:
+#      - '8095:8095'
+#    healthcheck:
+#      test: ['CMD', 'curl', '-f', 'http://localhost:8095/auth/health']
+#    environment:
+#      - SERVER_PORT=8095
+#      - SPRING_PROFILES_ACTIVE=dev
+#
+#  community-api:
```
chris-faulkner-mojdt, line 31: I think it's worth leaving comunity api and offender assessments in. I think I do use those sometimes. Can't you always start with `docker-compose up community-api` ?

chris-faulkner-mojdt (commented):
Can we also clear the .snyk file ? There's one suppression in there - maybe it can go ?

docker-compose.yml
jhevans: yeah this was me being lazy and not bothering to work out the proper way of doing it 😄  Will amend

chris-faulkner-mojdt (approved):
LGTM

PR #233, opened 2020-10-14
🚑  PIC-762: Fix default createdAfter filter to default to including all created after Unix Epoch

jhevans:


chris-faulkner-mojdt (approved):
LGTM

PR #259, opened 2020-11-12
✨ PIC-831 Add statusDate to response on breach for probation record

chris-faulkner-mojdt:


sarahyjja (approved):
I'm approving this PR as it's only adding a new entry 'statusDate' to the JSON 🤞

PR #261, opened 2020-11-17
✨ PIC-876 Add complex name to court case response.

chris-faulkner-mojdt:
Also made the CourtCase Mapper static to match others.


theDustRoom (approved):
LGTM

PR #263, opened 2020-11-20
✨ PIC-922: Add offenderId field to offenderDetail response

jhevans:


docker-compose.yml
```diff
@@ -4,7 +4,6 @@ services:
 
   postgres:
     image: circleci/postgres:11.6-alpine-ram
-    restart: always
```
jhevans, line 7: Removed these to default to "no" as it would restart the pods whenever I start my laptop regardless of whether I wanted them running.

chris-faulkner-mojdt (approved):
LGTM

PR #271, opened 2020-12-03
✨ PIC-933 Add call to get ProbationStatusDetail via new endpoint

chris-faulkner-mojdt:
Also includes breach and previouslyKnownTerminationDate


jhevans (approved):
LGTM 👍

PR #276, opened 2020-12-08
Pic 909 file size download

chris-faulkner-mojdt:
The document is streamed and you need to give spring / reactor a buffer size in the decoder. By default this is 262144 as given in AbstractDataBufferDecoder. So as to not change the value in other webclients, added a specific new one for the documents. Others will continue to work with the same values.


jhevans (approved):
👍

PR #281, opened 2020-12-10
🗃  PIC-981: Add missing indexes to court-case and offence tables

jhevans:


src/main/resources/db/migration/courtcase/V1120__add_index_to_offence_and_court_case.sql
```diff
@@ -0,0 +1,2 @@
+create index idx_offence_court_case_idx on offence (court_case_id);
+create index idx_court_case_court_code_case_no on court_case (court_code, case_no);
```
chris-faulkner-mojdt, line 2: We should already have an index on those two fields 
`CREATE UNIQUE INDEX court_case_court_case_uq ON COURT_CASE (COURT_CODE, CASE_NO);`

chris-faulkner-mojdt (approved):
LGTM

src/main/resources/db/migration/courtcase/V1120__add_index_to_offence_and_court_case.sql
jhevans: As discussed it turns out this got inadvertently removed in migration V1117 - just recording here for reference

PR #296, opened 2021-01-06
✨ PIC-905 Update probation status on current cases for same CRN

chris-faulkner-mojdt:
Also 

- changed OffenderMatchMapper so that it's static
- Created EntityHelper to reduce boiler plate / duplication of code in tests to create CourtCaseEntity in particular


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtCaseRepository.java
```diff
@@ -35,12 +32,14 @@
             nativeQuery=true)
     List<CourtCaseEntity> findByCourtCodeAndSessionStartTime(String courtCode, LocalDateTime sessionStartAfter, LocalDateTime sessionStartBefore, LocalDateTime createdAfter);
 
-    @Query(value = "Select cc FROM CourtCaseEntity cc "
-        + "WHERE cc.courtCode = :courtCode "
-        + "AND cc.sessionStartTime >= :start AND cc.sessionStartTime <= :end "
-        + "AND cc.caseNo NOT IN :caseNos")
-    List<CourtCaseEntity> findCourtCasesNotIn(@Param("courtCode") String courtCode,
```
chris-faulkner-mojdt, line 42: This was unused

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
```diff
@@ -40,7 +41,22 @@ public CourtCaseEntity createCase(String courtCode, String caseNo, CourtCaseEnti
                         },
                         () -> trackCreateEvents(updatedCase));
 
-        return courtCaseRepository.save(updatedCase);
+        return Mono.just(updatedCase)
+            .map(courtCaseRepository::save)
+            .doAfterTerminate(() -> updateStatusForCrn(updatedCase.getCrn(), updatedCase.getProbationStatus(), updatedCase.getCaseNo()));
```
chris-faulkner-mojdt, line 46: Reading of the API is that this should happen after the Mono has completed. However tests so far are unclear when. Should we consider a log debug statement on the save operation ?

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtCaseRepository.java
```diff
@@ -35,12 +32,14 @@
...
+    @Query(value="select cc.*, grouped_cases.min_created as first_created from court_case cc "
+        + "inner join (select min(created) as min_created, max(created) as max_created, case_no from court_case group_cc "
+        + "where crn = :crn "
+        + "and case_no != :caseNo "
+        + "and session_start_time >= current_date "
+        + "group by case_no, court_code) grouped_cases "
+        + "on cc.case_no = grouped_cases.case_no "
+        + "and cc.created = grouped_cases.max_created",
+        nativeQuery=true)
+    List<CourtCaseEntity> findOtherCurrentCasesByCrn(String crn, String caseNo);
```
jhevans, line 44: This doesn't seem to be making any distinction between the 'active' (i.e. the new case with the new probation status) case and the 'other' cases so presumably the 'active' case will be returned in the result set? Not sure if this is filtered down elsewhere but probably worth reflecting this in the method name either way and getting rid of the 'other'

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
jhevans: Yeah that would be good, I'd be tempted to create a telemetry event for it in fact 🤔  We'll get telemetry on the updates for each case though but might be useful to tie that back to a single probation status change.. but then again that's also covered by existing telemetry. No, maybe I've talked myself out of it, debug log is probably right! 😄

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
```diff
@@ -40,7 +41,22 @@ public CourtCaseEntity createCase(String courtCode, String caseNo, CourtCaseEnti
...
+    }
+
+    void updateStatusForCrn(String crn, String probationStatus, String caseNo) {
+        final var courtCases = courtCaseRepository.findOtherCurrentCasesByCrn(crn, caseNo)
+            .stream()
+            .filter(courtCaseEntity -> !courtCaseEntity.getProbationStatus().equalsIgnoreCase(probationStatus))
```
jhevans, line 52: Referring back to my earlier comment about the 'active' case being updated - this will make sure it isn't updated twice so happy for that to just be a name change.

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtCaseRepository.java
jhevans: ☝️ Ignore this I'm being an idiot - of course it's doing the other that's what `and case_no != :caseNo ` is for 🤦

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
jhevans: ☝️ Ignore this - previous comment wasn't valid

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderMatchesControllerIntTest.java
```diff
@@ -27,7 +28,7 @@
 @Sql(scripts = "classpath:after-test.sql", config = @SqlConfig(transactionMode = ISOLATED), executionPhase = AFTER_TEST_METHOD)
 public class OffenderMatchesControllerIntTest extends BaseIntTest {
 
-    private static final String COURT_CODE = "SHF";
+//    private static final String COURT_CODE = "SHF";
```
jhevans, line 31: This can go if it's not used

jhevans: Why did you decide to get rid of it as a matter of interest? I see it was used almost exclusively in String concatenations which does smell bit though I couldn't tell you why 😄

src/test/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/EntityHelper.java
```diff
@@ -0,0 +1,102 @@
+package uk.gov.justice.probation.courtcaseservice.jpa.entity;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.List;
+
+public class EntityHelper {
```
jhevans, line 7: Nice, good idea to create this 👍

jhevans (approved):
I think there was one commented line that could go but otherwise looks great.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
chris-faulkner-mojdt: log statement done.

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtCaseRepository.java
chris-faulkner-mojdt: Yes - I considered filtering this in the stream.

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderMatchesControllerIntTest.java
chris-faulkner-mojdt: Just created EntityHelper with a "real" looking court code and it filtered through to this - decided to standardise.

PR #299, opened 2021-01-07
PIC-604-added-rds-snapshot-files

katie-cawthorne:
This PR adds the required rds snapshot files to install the cronJobs.

`01-configmap-db-snapshot-script.yaml` - Defines the bash script used to perform the rds snapshots, and clean up of older ones
`02-cronjob.yaml` - Defines the cronjob which launches the job on a scheduled basis (see cron value in this file)
`03-prometheus-alert.yaml` - Define alert which will tell us if the snapshot job failed, alert sent to #probation_in_court_alerts
`rds-snapshot-job.yaml` -  Can be used to create one off snapshots on an adhoc basis.

I've added the required AWS IAM creds `access_key_id` and `secret_access_key` in our namespaces, this is required for the jobs to work.
I've also requested integration id for pingdom alerts to #probation_in_court_alerts slack channel but waiting for this to be actioned, once this is done we'll be able to run job 03-prometheus-alert.yaml to set the alerts up.


chris-faulkner-mojdt (approved):
LGTM

PR #320, opened 2021-02-03
✨ PIC-1078 telemetry for offender assessment call failures

chris-faulkner-mojdt:


src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderService.java
```diff
@@ -107,8 +109,10 @@ public ProbationRecord getProbationRecord(String crn, boolean applyDocumentFilte
                 probationRecord.setAssessment(findMostRecentByStatus(assessments).orElse(null));
             });
         } catch (OffenderNotFoundException e) {
+            telemetryService.trackTelemetryEvent(TelemetryEventType.GRACEFUL_DEGRADE);
```
jhevans, line 112: It would be good to include the specifics of what cause the degradation, could we send the info message as well? It's already logged obviously but saves us the hassle of having to correlate the two... in fact we could probably get rid of the log.warn() as this kinda supersedes it 🤔

jhevans (changes requested):
One suggestion, would be good to log some more info with the event

src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderService.java
jhevans: It's also possible that in future we'll have different degradation events so will need a way to differentiate

jhevans: In fact might be worth breaking it down a bit for easy querying, something along these lines:
```
{
  "description": "Excluding offenderAssessments", 
  "cause": exception,
  "crn": crn
}
```

chris-faulkner-mojdt: Given there is only one at the moment, it could only be for offender assessments. 
I followed the ticket and the maxim "do the simplest thing that works". I'll do it but stuff like this should be on the ticket at the outset.

chris-faulkner-mojdt: We already have an alert for the offender assessments failure so why do we have this telemetry ? I am a bit confused by the intended approach TBH

jhevans: Apologies, you're right I didn't make it clear on the ticket, will try to be more explicit in future.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderService.java
```diff
@@ -107,9 +109,9 @@ public ProbationRecord getProbationRecord(String crn, boolean applyDocumentFilte
                 probationRecord.setAssessment(findMostRecentByStatus(assessments).orElse(null));
             });
         } catch (OffenderNotFoundException e) {
-            log.info("assessment data missing from probation record (CRN '{}' not found in oasys)", crn);
```
chris-faulkner-mojdt, line 110: Removed these log statements as suggested. We'll still have the error level logging in the AssessmentRestClient

jhevans (approved):

PR #326, opened 2021-02-09
🐛 PIC-1102: Fix for session date/time updates not being reflected by …

jhevans:
…get cases endpoint

The get case list endpoint was filtering by session start time in the wrong SQL select statement meaning that for a given court_code/case_no combination where the case had been moved to a future date it would erroneously keep returning the case on the original date. 

The fix means that it should now only return a given case on the date given in the most recent update.


chris-faulkner-mojdt (approved):
LGTM

PR #334, opened 2021-02-18
✨ PIC-1107 change probation-status-detail endpoint

chris-faulkner-mojdt:
Behind a feature flag which is unset (apart from integration tests), change the way that probation-status-detail is collected. 

When community API updates are deployed, we can remove the old methods and the flag in code here. I have marked the methods to be removed as deprecated.


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ProbationStatus.java
```diff
@@ -4,7 +4,9 @@
 
 public enum ProbationStatus {
     CURRENT("Current"),
-    PREVIOUSLY_KNOWN("Previously known");
+    PREVIOUSLY_KNOWN("Previously known"),
+    NOT_SENTENCED("Not sentenced"),
```
jhevans, line 8: Apologies, I failed to mention on the ticket but we want to map `Not sentenced` to `No record` in the front end. I had to add `NOT_SENTENCED` into the community-api status so that the interface accurately reflects the reality (i.e. we do have a record, it's just they haven't been sentenced yet) but for the purposes of prepare-a-case we don't make a distinction between `No record` and `Not sentenced`

jhevans (changes requested):
One comment to fix something I missed on the ticket

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ProbationStatus.java
chris-faulkner-mojdt: Done - will return "no record" when serialising NOT_SENTENCED to Json

katie-cawthorne (approved):
LGTM

jhevans (approved):

PR #343, opened 2021-03-03
✨ PIC-1144 when CustodialStatus code is unknown - fix NPE

chris-faulkner-mojdt:
The incoming code was NOT_IN_CUSTODY which was not in our enum. I have added this but the new test is based on a code which is also not in the enum. No longer a null pointer.


src/test/resources/mocks/__files/offender-currentOrderHeaderDetail/GET_currentOrderHeaderDetail_NotInCustody_E396405.json
```diff
@@ -0,0 +1,16 @@
+{
+  "sentenceId": 1502924527,
+  "custodialType": {
+    "code": "XXXXX",
```
chris-faulkner-mojdt, line 4: This code is not in the CustodialStatus enum

katie-cawthorne (approved):
LGTM

PR #346, opened 2021-03-05
✨ PIC-1157: Always return manual updates for court-cases

jhevans:
I've had to approach this in a slightly weird way due to being unable to run the tests locally without my normal laptop but I've seen the test fail in CI and subsequently pass so I'm confident they're doing what they should.  I've done the most basic implementation and test I can given the difficulties I've got in making significant changes at the moment, happy to follow up with improvements when I'm back to my normal set up but would be good to merge asap to get PAC back in working order.


katie-cawthorne (approved):
LGTM

PR #351, opened 2021-03-11
✨ allow probation status to be null and handle "Possible nDelius record"

chris-faulkner-mojdt:
The CourtCaseResponse has a string for probation status and it has a getter method to determine if it is null / unmatched then can return "Possible nDelius record". So we don't store that value in the database anymore. PROBATION_STATUS can be null in these cases. Also is some migration. 
Fixed integration tests.


jhevans (approved):
👍

PR #354, opened 2021-03-17
✨ PIC-1197 return correct probation status

chris-faulkner-mojdt:
Also remove feature flag which allowed support for replaced community API calls

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -128,6 +128,3 @@ offender-service:
     # Comma-separated, all upper case
     included-statuses: "COMPLETE"
 
-feature:
-  flags:
-    use-community-api-for-probation-status: true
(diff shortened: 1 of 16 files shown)


src/main/resources/application.yml
```diff
@@ -127,3 +127,7 @@ offender-service:
   assessment:
     # Comma-separated, all upper case
     included-statuses: "COMPLETE"
+
+feature:
+  flags:
+    use-community-api-for-probation-status: true
```
jhevans, line 133: This can be removed now can't it?

jhevans, line 133:
```suggestion
```

jhevans (changes requested):
Think there's some config that can go, have suggested the changes in line so hopefully you can just do it in Github

chris-faulkner-mojdt:
> Think there's some config that can go, have suggested the changes in line so hopefully you can just do it in Github

Not sure how to do that

src/main/resources/application.yml
chris-faulkner-mojdt: Done

jhevans (approved):

PR #355, opened 2021-03-17
Pic 1172 fix offender detail probation status

jhevans:


src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/communityapi/model/CommunityApiOffenderResponse.java
```diff
@@ -26,7 +25,4 @@
     private String surname;
     private LocalDate dateOfBirth;
     private CommunityApiContactDetails contactDetails;
-    @JsonDeserialize(using = NumericBooleanDeserializer.class)
```
chris-faulkner-mojdt, line 29: The NumericBooleanDeserializer class is one of ours and is no longer used. We could leave it unused in case a similar class comes a long. Or get rid and re-institute when needed. What do you think ?

chris-faulkner-mojdt (approved):
One small comment about the NumericBooleanDeserialiser

PR #369, opened 2021-03-31
Pic 959 provider pact

chris-faulkner-mojdt:
Introduces PACT verification tests, implements State as needed by prepare a case consumer tests.
Verification runs against versions on the PACT broker.
Job to run verification against main of the consumer 

Other features of the change 
- removes Junit 4
- removes the unused `caseId` and `CRO` from `CourtCaseResponse`


.circleci/config.yml
```diff
@@ -175,6 +184,35 @@ jobs:
...
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - gradle-{{ checksum "build.gradle" }}
+            - gradle-
+      - run:
+          command: |
+            PACT_PROVIDER_VERSION="$CIRCLE_SHA1" \
+              PACT_PROVIDER_TAG="$CIRCLE_BRANCH" \
```
jhevans, line 203: Can we/should we use APP_VERSION for this so it matches the docker image tag?

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/mapper/CourtCaseResponseMapper.java
```diff
@@ -19,11 +19,9 @@
 
     public static CourtCaseResponse mapFrom(CourtCaseEntity courtCaseEntity, GroupedOffenderMatchesEntity groupedOffenderMatches) {
         return CourtCaseResponse.builder()
-                .caseId(courtCaseEntity.getCaseId())
```
jhevans, line 22: Have these just been dropped because we're not using them?

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/ProbationRecord.java
```diff
@@ -1,5 +1,6 @@
 package uk.gov.justice.probation.courtcaseservice.service.model;
 
+import java.util.List;
```
jhevans, line 3: Not sure how this has crept in twice 🤔

build.gradle
```diff
@@ -122,10 +149,6 @@ task smokeTest(type: Test) {
     }
```
jhevans, line 149: We can remove the smokeTest run config as well now

jhevans (changes requested):
A couple of very minor things. Otherwise looks good

.circleci/config.yml
chris-faulkner-mojdt: You mean for the PACT_PROVIDER_TAG ? I am not sure. As I understand it, this allows us to see what is happening on the branch as it is committed. 
Or are you talking about the PACT_PROVIDER_VERSION ?

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/mapper/CourtCaseResponseMapper.java
chris-faulkner-mojdt: No consumer is interested in them so, yes.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/ProbationRecord.java
chris-faulkner-mojdt: Weird - I have auto organise on for that. Done anyway.

build.gradle
chris-faulkner-mojdt: I'm also not sure about the whole WIP thing either. I think the correct way to deal with that is to use @Disabled on any tests which are in progress.

.circleci/config.yml
chris-faulkner-mojdt: Updated

jhevans (dismissed):

jhevans (approved):

PR #378, opened 2021-04-08
🛠 PIC-1134: Increase memory request

jhevans:


build.gradle
```diff
@@ -198,6 +198,10 @@ dependencies {
         because 'Transitive dependency. Previous versions have a known vulnerability'
     }
 
+    implementation ('com.nimbusds:oauth2-oidc-sdk:9.3.1') {
```
chris-faulkner-mojdt, line 201: have we tried excluding it  from spring-security ?

chris-faulkner-mojdt (dismissed):
One small comment to look at.

build.gradle
jhevans: I've not no but we do use oauth2 so seems likely it is used. Will have a go though and update as necessary

Funnily enough I was just thinking along similar lines and have an update to push regardless. I've gone back to check the other transitive dependencies to see if they've been patched and all but 1 of them (jetty-io) have been so I've removed those transitive overrides.

This might be a good practice to get into when we do find ourselves fettling vulnerabilities like this to check any that are already there that can be removed. I imagine it'll be a matter of days before spring patch it.

chris-faulkner-mojdt: Agreed. Maybe we could just note in the "because" phrase where it should be coming from and also agreed, it is very likely to be used so it's OK.

chris-faulkner-mojdt (dismissed):
LGTM

chris-faulkner-mojdt (approved):
LGTM

PR #384, opened 2021-04-15
Pic 1268 offender mgr email phone

chris-faulkner-mojdt:
Add the ADR for the decision about implementation.

Bulk of the change is in the way that the probation record is build. We now use a new endpoint to get offender managers. We had no use of the "offender all" endpoint here anymore so we lost the built-in exclusion checking so I added a call to the "basic" offender endpoint which does the same but we ignore the response body. 

Also split out the OffenderManager mapper code from the large-ish OffenderMapper.
The new endpoint in community API is live in prod now so no flag required


jhevans (approved):
Looks good to me 👍

PR #390, opened 2021-04-23
📈 PIC-1270: Log deployment message with version on application start

jhevans:
Not particularly happy with this as a solution but it works. Spring ought to log the version automatically from what I can tell but for reasons I don't understand it isn't. In the end I've taken the pragmatic approach and just logged it again as I managed to do that in less time than it took me debugging why it wasn't working in the first place 🙄

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ ext {
 
 def today = Instant.now()
 def todaysDate = LocalDate.now().format('yyyy-MM-dd')
-version = System.getenv('CI') ? "${todaysDate}.${System.getenv('CIRCLE_BUILD_NUM')}" : todaysDate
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -33,7 +33,7 @@ ext {
 
 def today = Instant.now()
 def todaysDate = LocalDate.now().format('yyyy-MM-dd')
-version = System.getenv('CI') ? "${todaysDate}.${System.getenv('CIRCLE_BUILD_NUM')}" : todaysDate
+def appVersion = System.getenv('CI') ? "${todaysDate}.${System.getenv('CIRCLE_BUILD_NUM')}" : todaysDate
```
jhevans, line 36: I discovered this was shadowing a Gradle variable hence renaming

chris-faulkner-mojdt (approved):
LGTM

PR #399, opened 2021-04-30
🚀 PIC-1348 reduce SQL statements to get offender match count

chris-faulkner-mojdt:
Each fetch for the count of matches was 2 SQL statements. This reduces it to 1


theDustRoom (approved):
Nice work 👍

PR #411, opened 2021-05-20
Pic 959 contract provider

chris-faulkner-mojdt:
Verifies and publishes PACT results to the broker. Only prepare-a-case at the moment. court-case-matcher PACTS are not being verified. We may need separate state verification provider classes. 

@jhevans - this doesn't line up 100% with the user-preferences implementation. It does not mix the 4.2.x dependencies with this one. Maybe we should review jointly.
`testImplementation("au.com.dius:pact-jvm-provider-junit5:4.0.10")`


jhevans:
The hmpps-user-preferences implementation is almost certainly crappier than this one as it's got that weird mixing going on so happy to merge this. I'll raise a ticket to review user-preferences and fix and hopefully will have a bit of time to sort it today or tomorrow but if not it'll be in the backlog ready.

jhevans (approved):

PR #432, opened 2021-06-07
✏️ PIC-1406: Update ADR and use default case list Last-Modified date

jhevans:
When going through the ADR it occurred to me that we can do better than just logging "now" as the last modified date and allow the client to cache if they want. As long as we consistently keep returning an arbitrary date in the past then it will be guaranteed to update as soon as a new update comes through (because the timestamp will be in the present not the past).

The date I chose is pretty arbitrary but I went for a nice round 2020-01-01 - so before we started getting real data but around the inception of the court-case-service as a thing.


chris-faulkner-mojdt (approved):
LGTM

PR #436, opened 2021-06-14
✨ PIC-1437 fetch requirements with convictions in probation record

chris-faulkner-mojdt:
Requirements are now returned on a per-conviction basis. Re-used code already found in the `ConvictionService`.


src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderService.java
```diff
@@ -119,6 +126,20 @@ public ProbationRecord getProbationRecord(String crn, boolean applyDocumentFilte
         return probationRecord;
     }
 
+    private Mono<Conviction> enrichConviction(String crn, Conviction conviction) {
+
+        var convictionId = Long.valueOf(conviction.getConvictionId());
+        var enrichedConvictionMono = Mono.zip(
```
jhevans, line 132: We only need to retrieve this lot for active convictions on the Probation Record screen I think, it'll show all the requirements on the order screen if they click through but for this page we only show them anything current.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderService.java
```diff
@@ -119,6 +126,20 @@ public ProbationRecord getProbationRecord(String crn, boolean applyDocumentFilte
...
+            offenderRestClient.getBreaches(crn, convictionId),
```
jhevans, line 133: I've just been reviewing the front end and I *think*  we can drop the breaches call for the probation record view entirely but will confirm. Sorry hadn't realised this when I wrote the ticket!

jhevans: Have just realised it did this before... I'm not sure why it can't just use the inBreach flag on the Conviction though as it only seems to toggle the breach badge on and off...

Never mind! It is redundant on this page but we've raised another ticket to remove the one place this is used  (the breach page) here https://dsdmoj.atlassian.net/browse/PIC-1442 and then another to follow up and delete here https://dsdmoj.atlassian.net/browse/PIC-1489

jhevans (changes requested):
Looks good but a minor change needed to only enrich active convictions

src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderService.java
chris-faulkner-mojdt: Yes - that's true and I asked Paul about that. I did wonder about CCS though generally. So in this area, it feels like we have an endpoint which is very specific for prepare a case (which it is :-) ) It returns active and inactive conviction requirements , only active PSS requirements and licence conditions. That starts to feel muddled. It's all so hardcoded to prepare a case. I started to feel uncomfortable.If we were doing something general, it would be active and inactive and let the client decide what's needed.

chris-faulkner-mojdt: Can do it, I just think we need to consider this for the future. We ought to document the responses around active / inactive so it's clear

jhevans: Yeah I know what you mean,  court-case-service is a bit confused between pure API and back-end for a front-end. Given our current use case is purely to serve the prepare-a-case UI I think that's probably ok for now, we ought not to let any other services call the probation record endpoint in particular though. If we do get any more consumers I think a bit of  ringfencing would be needed at the very least. 

FWIW the data *owned* by the court-case-service is pretty clean in the API it serves up. In that sense we've kinda got 2 different services rolled into 1.

jhevans (approved):

PR #440, opened 2021-06-15
🔧 PIC-1471 nginx cache to 25m

chris-faulkner-mojdt:


nginx/default.conf.template
```diff
@@ -1,4 +1,4 @@
-proxy_cache_path /var/cache/nginx levels=1:2 keys_zone=court_case_service_cache:10m max_size=10m
+proxy_cache_path /var/cache/nginx levels=1:2 keys_zone=court_case_service_cache:25m max_size=10m
```
jhevans, line 1: We probably also need to set the max_size which determines the usable disk space, looks like what you've changed there will only affect in-memory cache so not quite sure whether this would do it. There's more detail [here](https://www.nginx.com/blog/nginx-caching-guide/) 

Also I was expecting that we'd have some changes to the deployment.yaml to set the ephemeral storage limit and request, what did Cloud Platform have to say about this if anything?

jhevans (changes requested):

nginx/default.conf.template
chris-faulkner-mojdt: ask cloud platform replied with this 
https://mojdt.slack.com/archives/C57UPMZLY/p1623744301352800?thread_ts=1623679273.352100&cid=C57UPMZLY

Nothing to suggest.

chris-faulkner-mojdt: I am not sure how we set the storage limit on a pod but if 25M exceeded the default available I'd expected them to have said something. I guess we have to trust the testing done on the ticket you did.

jhevans: Ok cool, I suppose 25Mb is peanuts in cloud land so maybe not that surprising! Will approve now then

PR #467, opened 2021-07-16
📝  PIC-1135: Create guiding principles ADR

jhevans:
New ADR proposal for review and refinement


chris-faulkner-mojdt:
I think this is a really good use of the ADRs and more importantly a great statement of what CCS is about as an API. Great to have it stated. The one thing I wonder is whether, in addition to saying about the endpoints under /offender and such, we shouuld have some swagger tags to define what is general purpose and what is really gateway type thing for prepare-a-case only.  There's tags in community API to classify some of the endpoints.

jhevans:
Yeah I agree we should definitely start tagging the endpoints so there's a clear separation. If we decide what tags we want to use I'll add them to the ADR.

Could go with:
- `Domain`/`General purpose`
- `prepare-a-case`

🤔  I wonder if `Domain`, despite being technically more descriptive is needlessly obtuse and technical for this use-case and whether we'd be better off just going with `General purpose`. Might make more sense to a consumer.

chris-faulkner-mojdt (dismissed):
LGTM

sarahyjja (approved):
👍

PR #469, opened 2021-07-19
Pic 1520 add awaiting PSR and reports to probation-record

chris-faulkner-mojdt:
Note that the Conviction type used in the single conviction CCS endpoint but that one does not have the PSR reports populated.


src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderServiceTest.java
```diff
@@ -451,6 +455,38 @@ void givenBreachRequestFailsWithConnectionIssue_whenGetOffender_thenThrowExcepti
                 .isThrownBy(() -> service.getProbationRecord(CRN, true));
         }
 
+        @DisplayName("An offender which is awaiting PST will include PSR reports in the conviction")
```
jhevans, line 458: Typo 🤷

jhevans (approved):
LGTM, I found one typo which isn't really worth bothering with!

PR #472, opened 2021-07-21
Pic 1135 custodial order dates

jhevans:


src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/prisonapi/PrisonApiSentenceDetail.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.NoArgsConstructor;
+
+import java.time.ZonedDateTime;
+
+@Getter
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor(access = AccessLevel.PRIVATE, force = true)
+public class PrisonApiSentenceDetail {
+    private final ZonedDateTime homeDetentionCurfewActualDate;
```
chris-faulkner-mojdt, line 16: JUst a matter of interest, why ZonedDateTime ? Is this just to match what the prison API are doing ? If so, why do they do this ?

chris-faulkner-mojdt:
An overall comment - we decided to get the NOMS number as needed rather that capture and store it as part of the match process as we do with other fields. Is it worth slipping in an ADR for this ? Just a thought. If you think it's not worth it / of no value then that's fine.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/CustodyService.java
```diff
@@ -0,0 +1,68 @@
...
+import uk.gov.justice.probation.courtcaseservice.restclient.communityapi.model.OtherIds;
+import uk.gov.justice.probation.courtcaseservice.service.exceptions.CustodyNotFoundException;
+import uk.gov.justice.probation.courtcaseservice.service.exceptions.ExpectedCustodyNotFoundException;
+import uk.gov.justice.probation.courtcaseservice.service.exceptions.NomsNumberNotAvailableException;
+import uk.gov.justice.probation.courtcaseservice.service.model.Conviction;
+import uk.gov.justice.probation.courtcaseservice.service.model.Custody;
+import uk.gov.justice.probation.courtcaseservice.service.model.KeyValue;
+
+@AllArgsConstructor
+@RequestScope
```
chris-faulkner-mojdt, line 22: Just wondering why we have request scope here. I have used this before with the controller but not so familiar with using it on service tier. From what I read  , "The request scope creates a bean instance for a single HTTP request, while the session scope creates a bean instance for an HTTP Session."  See section 4 of this https://www.baeldung.com/spring-bean-scopes.

src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/prisonapi/PrisonApiSentenceDetail.java
jhevans: Yeah if you look at the docs [here](https://api-dev.prison.service.justice.gov.uk/swagger-ui/#/offenders/getOffenderSentenceDetails) you can see they include time and offset in the response but I'm told it's never used and only ever treated as a date. As to why - historical reasons! I think it was Andy M who told me that 🤔

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderControllerTest.java
```diff
@@ -51,6 +53,8 @@
     private ConvictionService convictionService;
     @Mock
     private BreachService breachService;
+    @Mock
+    private CustodyService custodyService;
```
chris-faulkner-mojdt, line 57: The bean is mocked but there is no test to cover the method

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderController_CustodyIntTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import uk.gov.justice.probation.courtcaseservice.BaseIntTest;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.equalTo;
+import static uk.gov.justice.probation.courtcaseservice.testUtil.TokenHelper.getToken;
+
+class OffenderController_CustodyIntTest extends BaseIntTest {
```
chris-faulkner-mojdt, line 13: Could possibly add an integration test for situation when the NOMS number is not found

jhevans:
> An overall comment - we decided to get the NOMS number as needed rather that capture and store it as part of the match process as we do with other fields. Is it worth slipping in an ADR for this ? Just a thought. If you think it's not worth it / of no value then that's fine.

Good thought, I'll put something together along those lines.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/CustodyService.java
jhevans: This is to allow the oauth2 to dynamically add the username to the community-api oauth token for its exclusion checks. There's an explanatory comment [here](https://github.com/ministryofjustice/court-case-service/blob/19164b1e6e5f6bff47a4f03a0bd8f96085d3d036/src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/OffenderRestClientFactory.java#L62) but the gist of it is that the Service has to be created per-request and injected with a community-api webclient which has been pre-seeded with the current username. 

Without the request scope it'll be instantiated once on the first request or at startup (can't recall which) and will from then on have an empty or incorrect value for username.

As to using it on the service tier, it shouldn't matter really where it's used as long as it's done on the same thread. Spring creates a proxy object which gets injected in place of the instance to the controller, so it' agnostic to the fact that the service is RequestScoped. 

It's quite a complex setup so I'm happy to walk you through the code at some point if you like, I should really get an ADR raised as well as it's decidedly non-obvious. I'll raise a task to do that.

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderController_CustodyIntTest.java
jhevans: That's what [this one](https://github.com/ministryofjustice/court-case-service/blob/d31a86f04c14d0bd458aa0a1ebba1ade3fc70ff1/src/test/java/uk/gov/justice/probation/courtcaseservice/controller/OffenderController_CustodyIntTest.java#L59) does I think. Is that what you mean?

chris-faulkner-mojdt (approved):
LGTM

chris-faulkner-mojdt (approved):
LGTM

PR #487, opened 2021-08-04
Pic 1571 remove current order header

jhevans:
Hold off on merging until [PIC-1569](https://dsdmoj.atlassian.net/browse/PIC-1569) is complete


src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/ConvictionRestClient.java
```diff
@@ -43,7 +44,7 @@
     private String convictionUrlTemplate;
 
     @Value("${community-api.current-order-header-url-template}")
```
chris-faulkner-mojdt, line 46: Should we rename the variable here ? current-order-header-url-template does not match the `sentenceStatusUrlTemplate`

chris-faulkner-mojdt (changes requested):

src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/ConvictionRestClient.java
jhevans: Good spot, will do

chris-faulkner-mojdt (approved):
LGTM

PR #488, opened 2021-08-05
📝 PIC-1586: Create stub endpoint for extended case

jhevans:


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedCourtCaseRequest.java
```diff
@@ -0,0 +1,23 @@
...
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE, force = true)
+@AllArgsConstructor
+public class ExtendedCourtCaseRequest {
+    private final String caseNo;
+    private final String courtCode;
+    private final String courtRoom;
+    private final List<HearingDay> hearingDays;
+    private final List<OffenceRequest> offences;
```
chris-faulkner-mojdt, line 20: In the new feed, the offences are a property of the defendant. They're not a direct part of the case, although we could model them this way with a reference to the defendant. 
None of this is really crucial because we are learning I suppose.

chris-faulkner-mojdt (dismissed):
One small question about the offences.

chris-faulkner-mojdt (approved):
LGTM

PR #491, opened 2021-08-06
✨ PIC-1581 add new JPA entities

chris-faulkner-mojdt:
Have also added some sample data. I tested the fetching of linked collections by debugging existing tests but didn't want to add methods to our services which would enable full integration tests.

As mentioned in a comment, the CourtCaseEntity cannot have 3 collections annotated with EAGER. We will need to look at this when we start fetching in our services but it looks like the Hibernate annotation LazyCollection could be a solution. We may need to start considering what should be lazy or not, and adjusting the session boundaries.

I haven't done it here but I think we should probably adopt a hearingDay field in the HEARING entity because we usually search on the hearing date (with no time) and it would give a more performant query.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ plugins {
     id 'org.springframework.boot' version '2.5.3'
     id 'io.spring.dependency-management' version '1.0.11.RELEASE'
     id 'com.github.ben-manes.versions' version '0.39.0'
-    id "org.flywaydb.flyway" version "7.12.0"
+    id "org.flywaydb.flyway" version "7.12.1"
     id 'jacoco'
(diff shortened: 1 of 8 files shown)


jhevans:
Good call on the hearing day, it might be easiest to do that now so that we can add the mapping in as part of the next story where we'll be populating all of the other fields anyway.

On a related note, this could have a significant impact on the performance of the court list query, when we come to change that over we should probably put it behind a toggle so we can back it out easily in the case that performance tanks

jhevans (approved):
Can we get that hearing day field added? Doesn't have to be as part of this PR but I agree it would be good to have

jhevans (approved):

PR #493, opened 2021-08-10
Pic 1582 populate new entities

chris-faulkner-mojdt:
The existing endpoints which receive only one defendant and one hearing date now also save that single hearing and defendant (with attached offences) to related JPA entities


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/CourtCaseRequest.java
```diff
@@ -56,13 +61,24 @@ public CourtCaseEntity asEntity() {
...
-                            .build();
+                        .sequenceNumber(i + 1)
+                        .offenceTitle(offence.getOffenceTitle())
+                        .offenceSummary(offence.getOffenceSummary())
+                        .act(offence.getAct())
+                        .build();
                 })
                 .collect(Collectors.toList());
+
+        final List<HearingEntity> hearings = List.of(HearingEntity.builder()
```
jhevans, line 72: Could be a singletonList for extra explicitness but List.of is fine

jhevans (approved):
LGTM 👍

PR #501, opened 2021-08-19
⚡️ PIC-1629: Add indexes to foreign key relationships to improve DB p…

jhevans:
…erformance


jhevans:
I've checked this in preprod and these indexes should deliver significant performance improvements. With regards to the other slow endpoints - I think that may have been resource contention rather than any inherently slow queries. As you mentioned @chris-faulkner-mojdt it doesn't look like any of the endpoints other than the PUT should have been affected.

chris-faulkner-mojdt (approved):
LGTM

PR #503, opened 2021-08-20
🐞 PIC-1609 create new endpoint controller for getCases (extended)

chris-faulkner-mojdt:
The  intent of the ticket was to remove the FK relationship from court_case to court and add it to court_case to hearing. This is done in the ticket. Also changed was the queries related to the endpoint /court/{court}/cases. As discussed, so as to enable more easy performance testing later, I have added new queries and a new controller at a new endpoint at /court/{court}/cases/extended. There is a new @Nested set of test classes in the CourtCaseControllerIntTest for this new endpoint. These test cases are exactly the same as the ones for the old get cases endpoint, save for the path.

The biggest part of the work was refactoring and adjustment of the before-test.sql. I had to add hearing records for most of the cases, as well as controlling the (usually generated) id fields for court_case. Most of these are now negative which I think makes analysis easier. I have added some look-up in a spreadsheet here. https://docs.google.com/spreadsheets/d/1o1wcn8AAMbCS84CRZ6tIcfb8DiuWMTNh1SOPc6Fje_o/edit#gid=0

Also, replaced the case_id values with UUIDs and made them unique where they should be,

Also, removed some unused SQL from the before-test.sql.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ import java.time.LocalDate
 
 plugins {
     id 'java'
-    id 'org.springframework.boot' version '2.5.3'
+    id 'org.springframework.boot' version '2.5.4'
     id 'io.spring.dependency-management' version '1.0.11.RELEASE'
     id 'com.github.ben-manes.versions' version '0.39.0'
     id "org.flywaydb.flyway" version "7.14.0"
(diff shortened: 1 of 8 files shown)


jhevans (approved):
LGTM 👍

PR #507, opened 2021-08-25
✨ PIC-1640 added a UUID to Defendant entity

chris-faulkner-mojdt:
Adds a UUID to the defendant. This has been done so that we have a natural key to link the offender matches to. It is expected that CCM will always send a value for the CP data. For the Libra feed, one should be generated in CCM and sent or else it will be generated in CCS.

### DIFF
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/DefendantEntity.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/DefendantEntity.java
@@ -23,6 +23,7 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 import lombok.ToString;
+import lombok.ToString.Exclude;
 import lombok.experimental.SuperBuilder;
 import org.hibernate.annotations.Type;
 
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/DefendantEntity.java
```diff
@@ -42,11 +43,16 @@
...
     private final Long id;
 
+    @ToString.Exclude
     @ManyToOne(optional = false)
     @JoinColumn(name = "COURT_CASE_ID", referencedColumnName = "id", nullable = false)
     @Setter
     private CourtCaseEntity courtCase;
 
+    @Column(name = "UUID", nullable = false)
+    private final String uuid;
```
jhevans, line 53: I'm torn as to whether `uuid` is a good name for this 🤔  I appreciate it is one but it feels wrong using the name of a data type where we probably ought to have a business concept. For example you wouldn't call a field `longId` or `stringId`. That said its business purpose is to provide a universally unique identifier... so maybe `uuid` *is* appropriate?

jhevans (approved):

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/DefendantEntity.java
chris-faulkner-mojdt: Yes - I thought the same as I was doing it. We use `caseId` in the `courtCaseEntity` to store the UUID, albeit that was partly because we were reusing an existing field. So maybe we should say `defendantId` ? The fact you mentioned it too means maybe we should do something.

PR #515, opened 2021-09-06
✨ PIC-1608 get case by case ID

chris-faulkner-mojdt:
Ticket refers to not returning the caseNo field which explains a fair amount of the change.

### DIFF
--- a/src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerIntTest.java
+++ b/src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerIntTest.java
@@ -1,6 +1,7 @@
 package uk.gov.justice.probation.courtcaseservice.controller;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.restassured.response.ValidatableResponse;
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerIntTest.java
```diff
@@ -373,6 +340,119 @@ void shouldReturnNotFoundForNonexistentCourt() {
...
+
+            assertThat(result.getDeveloperMessage()).contains("Case " + DELETED_CASE_ID + " not found");
+            assertThat(result.getUserMessage()).contains("Case " + DELETED_CASE_ID + " not found");
+            assertThat(result.getStatus()).isEqualTo(404);
+        }
+
+    }
+
+    private void validateResponse(ValidatableResponse validatableResponse, String startTime, String caseId) {
+        validatableResponse
```
jhevans, line 413: I quite like this, makes the test itself nice and concise whilst still doing the heavy lifting required 👍  Perhaps we should adopt this pattern more widely?

jhevans (approved):
LGTM 👍

PR #521, opened 2021-09-07
🔧 PIC-1642 clear out older records

chris-faulkner-mojdt:
Clear out everything before 1 June 2021. Copy same records to tables with `_archive` extension. 

We have to treat `offender_match` and `offender_match_group` differently because there is no FK relationship from `offender_match_group` to `court_case` and these tables do not have immutable records so we can't just delete based on a date. Need to delete parent records from `court_case` and then delete orphans from `offender_match` based on case no and court code.

I have put the SQL In a script to be executed by flyway. We could arguably do this manually with more control, and checking as we go. Happy to reconsider the approach.


jhevans (approved):
LGTM 👍

PR #524, opened 2021-09-09
✅ PIC-1665 PR comments - improve handling of multiple cases

chris-faulkner-mojdt:
Now using hearing collection to set court room / court code etc in the `CourtCaseResponse`. Also added a `hearingDays` array to `CourtCaseResponse`.

### DIFF
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/mapper/CourtCaseResponseMapper.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/mapper/CourtCaseResponseMapper.java
@@ -65,7 +65,7 @@ public class CourtCaseResponseMapper {
 
         var targetHearing = hearings
             .stream()
-            .filter(hearingEntity -> hearingEntity.getHearingDay().isEqual(Optional.ofNullable(hearingDate).orElse(LocalDate.MIN)))
+            .filter(hearingEntity -> hearingDate == null ? true : hearingDate.isEqual(hearingEntity.getHearingDay()))
             .findFirst();
 
         // Populate the top level fields with the details from the single hearing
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/mapper/CourtCaseResponseMapper.java
```diff
@@ -47,16 +46,60 @@ public static CourtCaseResponse mapFrom(CourtCaseEntity courtCaseEntity, int mat
...
+        return mapFrom(courtCaseEntity, matchCount, true, hearingDate);
     }
 
+    static void buildHearings(CourtCaseResponseBuilder builder, CourtCaseEntity courtCaseEntity, LocalDate hearingDate) {
+        var hearings = Optional.ofNullable(courtCaseEntity.getHearings())
+            .orElse(Collections.emptyList());
+
+        var targetHearing = hearings
+            .stream()
+            .filter(hearingEntity -> hearingEntity.getHearingDay().isEqual(Optional.ofNullable(hearingDate).orElse(LocalDate.MIN)))
```
jhevans, line 68: Could use a map here to move some stuff off this line. 

Also the use of the LocalDate.MIN is a bit odd and it's not immediately obvious what its significance is. If I'm reading it correctly then we only want to filter if there is a hearingDate and otherwise just get the first. That being the case I think this would be equivalent and a bit more concise:
```suggestion
            .map(HearingEntity::getHearingDay)
            .filter(hearingDay-> hearingDate == null ? true : hearingDay.isEqual(hearingDate)))
```

jhevans (dismissed):
One minor readability suggestion but looks good

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/mapper/CourtCaseResponseMapper.java
chris-faulkner-mojdt: I can't use a map because I am after the actual hearing and not the date. 
I always usd to prefer ternary operator in cases like this but was deterred by a comment on a community-api when I used it and was directed to use Optional.ofNullable because it's "more functional". I am still not convinced on that because Optional.ofNullable is more verbose, often less clear (to me at least) and results in an extra object being created.

So I have not used the map but have added the ternary and prefer it :-)

theDustRoom (approved):
LGTM

PR #528, opened 2021-09-15
✨ PIC-1619 - save offender match group by defendant and case id

chris-faulkner-mojdt:
1. Allows for the saving of offender matches by case and defendant ID.
2. Adds case_id and defendant_id to the offender_match_group table, makes a compound index of these 2 fields and that is unique.
3. Migrates existing case and defendant IDs. There is a known issue whereby some latest case versions do not have defendant IDs. It has been  decided on slack that we will work around that in subsequent tickets for match detail and match counts by falling back to looking up defendant id from court code and case no
4. We will investigate further the source of the issue of missing defendant IDs

### DIFF
--- a/src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderMatchServiceTest.java
+++ b/src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderMatchServiceTest.java
@@ -1,9 +1,12 @@
 package uk.gov.justice.probation.courtcaseservice.service;
 
 
+import lombok.Data;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentMatcher;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import reactor.core.publisher.Mono;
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/justice/probation/courtcaseservice/service/OffenderMatchServiceTest.java
```diff
@@ -50,212 +54,290 @@
...
         service = new OffenderMatchService(courtCaseService, offenderMatchRepository, offenderRestClientFactory);
     }
 
-    @Test
-    void givenValidRequest_whenCreateGroupedMatchesCalled_thenCreateAndReturnMatch() {
-        when(offenderMatchRepository.findByCourtCodeAndCaseNo(COURT_CODE, CASE_NO)).thenReturn(Optional.empty());
-        when(courtCaseService.getCaseByCaseNumber(COURT_CODE, CASE_NO)).thenReturn(courtCaseEntity);
-        when(offenderMatchRepository.save(any(GroupedOffenderMatchesEntity.class))).thenReturn(groupedOffenderMatchesEntity);
+    @Nested
+    class CreateOrUpdateByCourtAndCaseNo {
```
jhevans, line 72: I've got to say I'm not sure I'm a fan of nested test classes, I find it difficult to track the different methods and variables in scope and have found myself inadvertently editing the wrong data/methods etc as a result of multiple nested tests shadowing each other.  Maybe it's just lack of familiarity on my part, not suggesting you change it, just something I've been mulling 🤔

jhevans (approved):
LGTM 👍

PR #531, opened 2021-09-16
✨ PIC-1694 duplicate cases in the case list 

chris-faulkner-mojdt:
When there is more than one defendant on the case, then we duplicate the case in the case list response. The case list response remains the same so that no new work is required in the front end.


jhevans (approved):
LGTM 👍

PR #547, opened 2021-10-06
Pic 1752 pact fix

chris-faulkner-mojdt:
PACTs to prepare a case updates. Most of the grunt work was sorting the before-test.sql. There's still work to do in there


jhevans (approved):
Needs main merging in but happy to approve. Let me know if you need any help with the merge.

PR #554, opened 2021-10-13
🐞 PIC-1791 - update other probation status just for most recent version

ghost:
I think this may also fix 1696 - see change to the mapper.

Initially, we started with `findOtherCurrentCasesByCrnNotCaseId` . This was selecting all versions of  a case rather than the most recent. 

The fix also required me to update the integration test data so that it's realistic and uses hearing / defendant entries.

### DIFF
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtCaseRepository.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtCaseRepository.java
@@ -149,11 +149,20 @@ public interface CourtCaseRepository extends CrudRepository<CourtCaseEntity, Lon
             nativeQuery= true)
     List<CourtCaseEntity> findOtherCurrentCasesByCrn(String crn, String caseNo);
 
-    @Query(value = "SELECT cc.*, null as first_created FROM court_case cc "
-        + "where cc.case_id != :caseId "
-        + "and cc.session_start_time >= current_date "
-        + "and cc.crn = :crn "
-        + "and cc.deleted = false ",
+    @Query(value = "select cc.*, grouped_cases.min_created as first_created "
+        + "        from court_case cc "
+        + "           inner join "
+        + "               (select min(group_cc.created) as min_created, max(group_cc.created) as max_created, group_cc.case_id from court_case group_cc "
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/CourtCaseRepository.java
```diff
@@ -149,11 +149,20 @@
...
+        + "        from court_case cc "
+        + "           inner join "
+        + "               (select min(group_cc.created) as min_created, max(group_cc.created) as max_created, group_cc.case_id from court_case group_cc "
+        + "                   inner join hearing on hearing.court_case_id = group_cc.id  "
+        + "                   group by group_cc.case_id) grouped_cases "
+        + "               on cc.case_id = grouped_cases.case_id "
+        + "           inner join hearing h on h.court_case_id = cc.id "
+        + "           inner join defendant d on d.court_case_id = cc.id "
+        + "        where h.hearing_day >= CURRENT_DATE"
+        + "        and d.crn = :crn"
```
chris-faulkner-mojdt, line 162: We are going to have to a lot of this in the clean up. We can't look for CRN in court_case.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
```diff
@@ -142,7 +146,7 @@ void updateOtherProbationStatusForCrnByCaseId(String crn, String probationStatus
         if (crn != null) {
             final var courtCases = courtCaseRepository.findOtherCurrentCasesByCrnNotCaseId(crn, caseId)
                     .stream()
-                    .filter(courtCaseEntity -> !courtCaseEntity.getProbationStatus().equalsIgnoreCase(probationStatus))
+                    .filter(courtCaseEntity -> hasAnyDefendantsSameCrnDifferentProbationStatus(courtCaseEntity.getDefendants(), crn, probationStatus))
```
chris-faulkner-mojdt, line 149: Can't simply look in the court case entry for CRN. We need to look to see if any of the defendants have a CRN which matches but does NOT match the probation status.

jhevans (approved):
👍  Nice, that seems like some really tricky testing but it reads well

PR #582, opened 2021-11-09
Pic 1610 - Remove hearing data from CourtCaseEntity

jhevans:


chris-faulkner-mojdt (commented):
Need to update R_seed_data to remove unused fields

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedCourtCaseRequestResponse.java
```diff
@@ -45,11 +46,26 @@
     @NotEmpty
     private final List<Defendant> defendants;
 
-    public static ExtendedCourtCaseRequestResponse of(CourtCaseEntity courtCase) {
+    public static ExtendedCourtCaseRequestResponse of(CourtCaseEntity courtCase, boolean throwMultipleCourtsException) {
         return ExtendedCourtCaseRequestResponse.builder()
                 .caseNo(courtCase.getCaseNo())
                 .caseId(courtCase.getCaseId())
-                .courtCode(courtCase.getCourtCode())
+                .courtCode(Optional.ofNullable(courtCase.getHearings())
```
chris-faulkner-mojdt, line 53: Why are we keeping the courtCode at the top level here ? I don't think prepare-a-case uses this object.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
```diff
@@ -184,7 +183,7 @@ public CourtCaseEntity getCaseByCaseIdAndDefendantId(String caseId, String defen
     }
 
     private void validateEntity(String caseId, CourtCaseEntity updatedCase) {
-        checkCourtExists(updatedCase.getCourtCode());
+        checkCourtExists(updatedCase.getHearings().get(0).getCourtCode());
```
chris-faulkner-mojdt, line 186: I think we should validate all the courts here, not just the first. Make a distinct list and validate?

chris-faulkner-mojdt (commented):
Great to see some repo methods going ! Would like to go through this multiple court codes issue

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedCourtCaseRequestResponse.java
```diff
@@ -45,11 +46,26 @@
...
+                        .map(hearings -> hearings.stream()
+                                .map(HearingEntity::getCourtCode)
+                                .distinct()
+                                .collect(Collectors.toList()))
+                        .flatMap(distinctCodes -> {
+                            if (distinctCodes.size() > 1) {
+                                final var message = String.format("ExtendedCourtCaseResponse.courtCode could have multiple possible values %s", String.join(",", distinctCodes));
```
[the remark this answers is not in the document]
jhevans: Technically I think you're right but as a result of how we unpack CP payloads I don't think it can actually happen (see below for why). Also it's key to the structure of this response that we have only one `courtCode` as it's a top level field applied across all hearings (which also define a courtCode which is a bit odd really), arguably we should remove that field in fact coming to think of it but that would require changes to court-case-matcher 🤔  What I wanted to avoid was implying to the caller that the courtCode provided at the top level was applicable to the entire payload when in fact there are multiple courts across the different hearings. That's the `IllegalState` so to speak. 

I am now wondering whether the correct approach here is to deprecate that top level courtCode field entirely though 🤔  What do you reckon?

Re the reality of multiple court codes - We get our court code from the  [courtCentre](https://github.com/ministryofjustice/court-case-matcher/blob/dbcc835f60489346354ffed0418f3af0e1d8c534/src/main/java/uk/gov/justice/probation/courtcasematcher/messaging/model/commonplatform/CPHearing.java#L44) which occurs once at the hearing level. Each hearing then has  this applied despite the fact they could strictly speaking have different `courtCentreIds`. It's a bit weird and I was tempted to just assume that it will always have a single court but eventually decided that a bit of defensive coding wouldn't hurt.

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedCourtCaseRequestResponse.java
jhevans: Does it not? in which case we should drop it. I thought the matcher did as well but have just checked and it doesn't. Will remove.

chris-faulkner-mojdt (approved):
LGTM

PR #587, opened 2021-11-11
Pic 1859 save gender

chris-faulkner-mojdt:
We will now store "sex" values as they are supplied by common platform, i.e. 

`    "gender": {
      "description": "The person gender",
      "type": "string",
      "enum": [
        "MALE",
        "FEMALE",
        "NOT_KNOWN",
        "NOT_SPECIFIED"
      ]
    },`

This change encompasses the following

1. migrates existing values in the database which from Libra are 'M', 'F' and 'N'.
2. CP always sends a value so we will model the field as not-null
3. Use an enum in CourtCaseEntity, to ensure correct values saved
4. CourtCaseRequest - as sent by prepare-a-case requires values of 'M', 'F' etc and so we don't have to change that client, the defendant sex is a string
5. CourtCaseResponse  - defendant sex is value from the enum
6. ExtendedCourtCaseRequestResponse - using a string because the CCM pacts depend on it. Need to raise a ticket in CCM to send 'MALE','FEMALE', etc as they are received from CP


jhevans (approved):

jhevans:
Looks good subject to PACT failures

PR #589, opened 2021-11-12
✨ PIC-1859 defendant sex may be null when received

chris-faulkner-mojdt:
Only line of prod code is to remove an annotation, allowing defendant sex to be null (can happen from LIBRA), the rest is tests to cover


theDustRoom (approved):
LGTM

PR #597, opened 2021-11-18
✨ PIC-1812 add offender entity

chris-faulkner-mojdt:
Highlights

1. New OffenderEntity which duplicates fields from DefendantEntity. These are populated on save and if present, returned in responses to various requests for court cases, either individually or in the case list. Features of the relationship
- one way relationship from defendant to offender (many defendants to one offender
- CRN unique key in offender table
- foreign key from defendant to offender

2. Updates to TelemetryService. Events were still present based on linking and unlinking of defendants to CRN based on the CRN found in the court case. These possibly ought to have been moved with the original refactor to allow multiple defendants. We still have CRN in court case table and this needs to go.
3. Similar to TelemetryService, the group offender match was looking at the CRN on a case rather than the defendant ```boolean crnMatches = match.getCrn().equals(updatedCase.getCrn());```
3. ADR write-up
4. Removal of code which saved values of probation status, breach, etc on  cases/ defendants with the same CRN.

Notes 
1. As posted on Slack. Some of the PACTs for prepare-a-case specify "Possible NDelius record" and then also have values for awaiting PSR and so on which makes no sense logically. I have "fixed" this by making the integration test have no offender so you get "Possible NDelius record" but then falls back to the defendant-based fields for AWAITING_PSR. This will break again when we remove the defendant-based fields. So we need a ticket to change the values expected by prepare-a-case. Probably just make the probation status CURRENT or something.


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/mapper/CourtCaseResponseMapper.java
```diff
@@ -165,4 +134,27 @@ private static void addDefendantFields(CourtCaseResponseBuilder builder, Defenda
         // Offences
         builder.offences(mapOffencesFromDefendantOffences(defendantEntity.getOffences()));
     }
+
+    private static void addOffenderFields(CourtCaseResponseBuilder builder, DefendantEntity defendant, OffenderEntity offender) {
+        builder
+            .awaitingPsr(Optional.ofNullable(offender)
+                                    .map(OffenderEntity::getAwaitingPsr)
+                                    .orElse(defendant.getAwaitingPsr()))
```
chris-faulkner-mojdt, line 142: This fall back to defendant for offender based fields is only here to allow PACTs to continue to work. See https://dsdmoj.atlassian.net/browse/PIC-1907 which should be done as we remove the defendant fields

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
```diff
@@ -37,53 +38,55 @@
...
                 .ifPresentOrElse(
-                        (existingCase) -> {
+                        existingCase -> {
                             updatedCase.getDefendants()
-                                    .forEach(defendantEntity ->
-                                            updateOffenderMatches(existingCase, updatedCase, defendantEntity.getDefendantId())
-                                    );
-                            trackUpdateEvents(updatedCase, existingCase);
+                                    .forEach(defendantEntity -> updateOffenderMatches(existingCase, updatedCase, defendantEntity.getDefendantId()));
+                            trackUpdateEvents(existingCase, updatedCase);
```
chris-faulkner-mojdt, line 66: Yes - I switched the order of parameters. I noticed that confirmAndRejectMatches took existingCase first and updatedCase as second parameter and decided to make these other methods do the same.

src/main/resources/db/migration/courtcase/V2105__add_offender.sql
```diff
@@ -0,0 +1,27 @@
...
+        AWAITING_PSR                        BOOLEAN,
+        CREATED                             TIMESTAMP NOT NULL DEFAULT now(),
+        CREATED_BY                          TEXT,
+        LAST_UPDATED						TIMESTAMP,
+        LAST_UPDATED_BY						TEXT,
+        VERSION 							int4 NOT NULL DEFAULT 0,
+        DELETED                             BOOLEAN NOT NULL DEFAULT FALSE
+    );
+
+    insert into OFFENDER ( CRN, PREVIOUSLY_KNOWN_TERMINATION_DATE, PROBATION_STATUS, SUSPENDED_SENTENCE_ORDER, BREACH, PRE_SENTENCE_ACTIVITY, AWAITING_PSR, CREATED_BY)
```
chris-faulkner-mojdt, line 20: Have tried this out in dev and it seems OK - worth doing again in pre-prod

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerIntTest.java
```diff
@@ -505,14 +528,13 @@ private void validateResponse(ValidatableResponse validatableResponse, String st
             .body("offences[0].offenceSummary", equalTo("On 01/01/2015 at own, stole article, to the value of £987.00, belonging to person."))
             .body("offences[0].act", equalTo("Contrary to section 1(1) and 7 of the Theft Act 1968."))
             .body("offences[1].offenceTitle", equalTo("Theft from a different shop"))
-            .body("probationStatus", equalTo(PROBATION_STATUS))
-            .body("probationStatusActual", equalTo(null))
-            .body("previouslyKnownTerminationDate", equalTo(LocalDate.of(2010, 1, 1).format(DateTimeFormatter.ISO_LOCAL_DATE)))
-            .body("suspendedSentenceOrder", equalTo(true))
-            .body("breach", equalTo(true))
+            .body("probationStatus", equalTo("Possible NDelius record"))
+//            .body("probationStatusActual", equalTo("CURRENT"))
```
jhevans, line 532: Think this probably wasn't meant to be commented out - will uncomment and fix in a follow up commit

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/model/CourtCaseRequestTest.java
```diff
@@ -171,5 +180,6 @@ void givenOptionalFieldsNotPresent_whenCallAsEntity_returnCourtCaseEntity() {
         assertThat(entity.getDefendants().get(0).getDefendantId()).isNotNull();
         assertThat(entity.getDefendants().get(0).getDefendantName()).isEqualTo(DEFENDANT_NAME);
         assertThat(entity.getDefendants().get(0).getDateOfBirth()).isEqualTo(DEFENDANT_DOB);
+//        assertThat(entity.getDefendants().get(0).getOffender()).isNull();
```
jhevans, line 183: Probably shouldn't be commented, will amend

jhevans (dismissed):

PR #611, opened 2021-11-30
🗃 PIC-1615: Prevent optimistic locking failures for concurrent updates

jhevans:
Having examined the code I'm pretty sure I can see where the issues are arising and it seems that we need a slightly higher level of isolation than the default (`Isolation.READ_COMMITTED`) to achieve this. Here's the documentation for `Isolation.REPEATABLE_READ` as its non-obvious. The issue I believe we're seeing is in bold italics:

> A constant indicating that dirty reads and non-repeatable reads are prevented; phantom reads can occur. This level prohibits a transaction from reading a row with uncommitted changes in it, and **_it also prohibits the situation where one transaction reads a row, a second transaction alters the row, and the first transaction rereads the row, getting different values the second time_** (a "non-repeatable read").


markberridge (approved):
LGTM

PR #633, opened 2021-12-16
🔉 PIC-1946: Correct log message

jhevans:
close method is also called on success so current message is misleading


theDustRoom (dismissed):
LGTM. Not sure why the build failed.

jhevans:
I forgot there was a test for that 🙄

PR #657, opened 2022-01-17
🔧 PIC-1967: Use a very long time ago as default createdAfter date

jhevans:


jhevans:
Not to be merged just yet - pretty sure it's good but want to run it past everyone in standup in case there's a good reason why it's like it is.

PR #663, opened 2022-01-24
Pic 1864

rschallapalli:


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CourtCaseEntity.java
```diff
@@ -158,12 +85,6 @@
     @Column(name = "manual_update", nullable = false, updatable = false)
     private boolean manualUpdate;
 
-    @Deprecated(forRemoval = true)
-    @PrePersist
-    public void isManualUpdate(){
```
jhevans, line 163: Was this used anywhere do you know? I presume not or it would have broken stuff, happy for it to remain deleted as long as we're confident.

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/OffenceEntity.java
```diff
@@ -40,13 +40,6 @@
     @JsonIgnore
     private final Long id;
```
jhevans, line 41: I think we ought to be able to delete this entity as well though there is [another ticket ](https://dsdmoj.atlassian.net/browse/PIC-1795)to pick that up... not sure why we've split it into multiple tickets! Up to you how you want to do it.

jhevans (approved):
A couple of comments for review but happy to merge

PR #676, opened 2022-02-04
PIC-2006: Remove listNo required constraint on hearing.listNo

rschallapalli:


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/ListNoValidator.java
```diff
@@ -0,0 +1,80 @@
...
+    private void validateListNo(ExtendedCourtCaseRequestResponse courtCase) {
+
+        List<HearingDay> hearingDays = Optional.ofNullable(courtCase.getHearingDays())
+                .orElse(Collections.emptyList());
+
+        var hearingDaysWitListNo = hearingDays.stream().filter(hearingDay -> StringUtils.isNotBlank(hearingDay.getListNo())).count();
+
+        if(hearingDaysWitListNo > 0) {
+            if(hearingDaysWitListNo != hearingDays.size()) {
+                throw new ConflictingInputException("listNo is missing from one or more hearingDays[]");
```
jhevans, line 58: This probably isn't the right exception for this, in fact I think ConstraintValidators shouldn't throw at all but just return false if the field isn't valid. The error message needs to be passed to the ConstraintValidatorContext I think. Some examples [here](https://docs.oracle.com/javaee/7/api/javax/validation/ConstraintValidatorContext.html#buildConstraintViolationWithTemplate-java.lang.String-)

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/ValidateListNo.java
```diff
@@ -0,0 +1,22 @@
...
+
+import static java.lang.annotation.ElementType.*;
+
+
+/**
+ * Custom annotation to provision validation of listNo field.
+ */
+@Target({TYPE, ANNOTATION_TYPE, PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+@Constraint(validatedBy = ListNoValidator.class) // 1.
```
jhevans, line 17: // 1 etc are redundant

jhevans (commented):
Couple more changes needed to the validator to handle validation idiomatically

jhevans (approved):

PR #701, opened 2022-02-23
PIC-1945 deployment failure ix 2

rschallapalli:


data_dictionary/Dockerfile
```diff
@@ -1,11 +1,12 @@
...
+FROM nginx:1.21.3-alpine
 
 RUN chown -R nginx:nginx /etc/nginx
 RUN chown -R nginx:nginx /var/cache/nginx
 RUN chown -R nginx:nginx /var/log/nginx
+RUN touch /var/run/nginx.pid && chown -R nginx:nginx /var/run/nginx.pid
 
 RUN apk -U upgrade
 
+EXPOSE 8080
```
jhevans, line 10: Should this not be 80? If it's working then we may not need this at all 🤔

jhevans (approved):

PR #710, opened 2022-03-02
PIC-2045 add missing meata data to data extractor job

rschallapalli:


helm_deploy/values-dev.yaml
```diff
@@ -19,6 +19,10 @@ ingress:
     - host: court-case-service-dev.apps.live-1.cloud-platform.service.justice.gov.uk
   path: /
 
+analytics_data_extractor:
```
jhevans, line 22: If this is constant across environments you could add it to values-live.yaml

PR #787, opened 2022-05-06
PIC-1438 add aliases to offender match

rschallapalli:


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/OffenderAliasEntity.java
```diff
@@ -0,0 +1,23 @@
...
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE, force = true)
+@AllArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class OffenderAliasEntity {
+
+    private final LocalDate dateOfBirth;
+    private final String firstName;
+    private final String surname;
+    private final String gender;
```
rschallapalli, line 22: Do we have full list of values for this as the values are different in the DefendantEntity where the gender is referred to as sex and the values are "MALE", "FEMALE", "NOT_KNOWN", "NOT_SPECIFIED" should we create a new enum for gender to accept Male, Female.. ? @jhevans

jhevans, line 22: Looking at the API [here](https://probation-offender-search-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/offender-match/match) it looks like an array of `middleNames` can be returned as well so we should add those...

It would be good to reuse the `NamePropertiesEntity` in fact... but that stores it as three forenames 🙄 
```
    @Type(type = "jsonb")
    @Column(columnDefinition = "jsonb", name = "NAME", nullable = false)
    private final NamePropertiesEntity name;
    ...
    private String forename1;
    private String forename2;
    private String forename3;
```

That's annoying... We'll just have to get the matcher to mung any excess middle names down into forename3 when it posts.  

EDIT: Don't do this ☝️  See this 👇

jhevans: Urgh... but then again this whole class is getting the JSONB treatment... so we're kinda making a rod for our own back if we do that.  

Ok change of plan - Keep it as we receive it from offender search, and store it exactly as specified with a List of `middleNames`. The matches tables are effectively a local cache of offender search results anyway, it's not data that we 'own' as such so I think perhaps our interfaces should default to whatever offender search produces.  I'll add an ADR to this effect so we've got some guiding principles for future questions around this.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/MatchType.java
```diff
@@ -5,6 +5,7 @@ public enum MatchType {
     NAME_DOB_PNC,
     /** Matches to all the parameters supplied but at least one from the one of the aliases associated to the offender */
     NAME_DOB_ALIAS,
+    ALL_SUPPLIED_ALIAS,
```
jhevans, line 8: We can use the existing NAME_DOB_ALIAS for these records - there's technical reasons why we have to map the value returned by Offender Search to these values. I can't remember the specifics off the top of my head but will refresh my memory and update you when I've done so, or put it on the matcher ticket.

jhevans: The code that does the mapping is [here](https://github.com/ministryofjustice/court-case-matcher/blob/main/src/main/java/uk/gov/justice/probation/courtcasematcher/restclient/model/offendersearch/OffenderSearchMatchType.java)

So the reason we do that mapping is because the value that Offender Search returns `ALL_SUPPLIED` doesn't tell us what it was that we supplied, e.g. name, dob pnc. It matters to us whether we've matched on PNC as this is considered to be an authoritative match - if it's just name and dob then that's not as high quality so we introduced this mapping to retain that information.

jhevans (changes requested):
A middleNames field is need that was missed off the ticket, apologies!

vb-moj (commented):
testing

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/OffenderAliasEntity.java
rschallapalli: addressed

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/MatchType.java
rschallapalli: addressed

vb-moj (approved):

PR #806, opened 2022-05-24
PIC-2124 fix duplicate court room numbers in court room filter

rschallapalli:


jhevans (changes requested):
NPE causing the Pact tests to fail

jhevans (approved):

PR #814, opened 2022-05-30
Pic 2167

vb-moj:


src/main/java/uk/gov/justice/probation/courtcaseservice/service/mapper/OffenderMatchMapper.java
```diff
@@ -32,6 +32,15 @@ public static GroupedOffenderMatchesEntity newGroupedMatchesOf(String defendantI
         return group;
     }
 
+    public static GroupedOffenderMatchesEntity newGroupedMatchesOf(String defendantId, GroupedOffenderMatchesRequest offenderMatches, String caseId) {
```
vb-moj, line 35: This is almost the same as the existing method. Instead of a full case, we just pass the caseId.

rschallapalli (approved):

PR #820, opened 2022-06-08
Pic 2197

rschallapalli:


src/main/java/uk/gov/justice/probation/courtcaseservice/restclient/OffenderRestClient.java
```diff
@@ -125,9 +125,8 @@ public Mono<List<Breach>> getBreaches(String crn, Long convictionId) {
                 .retrieve()
                 .onStatus(HttpStatus::is4xxClientError, resp -> clientHelper.handleConvictionError(crn, convictionId, resp))
                 .bodyToMono(CommunityApiNsiResponse.class)
-                .onErrorMap(e1 -> {
+                .doOnError(e1 -> {
```
jhevans, line 128: This will swallow the error rather than causing a failure at the calling method, is that intentional?

jhevans: Sorry! Ignore me! Do onError is right as you say, they're actually functionally equivalent!

jhevans (approved):

jhevans (approved):

PR #830, opened 2022-06-14
💥PIC-2093: [skip ci] Sketch out possible restructure for discussion

jhevans:


jhevans:
I've created this draft PR as a straw man demonstration of a restructure we could do to achieve the benefits of our current immutable implementation whilst reverting to pure JPA/Spring Data for relationship management. The hope is this will be much more readable for anyone who's not intimately familiar with the peculiarities and history of the service whilst giving us a full version history for all important changes to a case as will be required for case tracking when (and if) it comes around.  

It would take a significant effort to implement but long term it could be simpler and more maintainable.  Please throw stones at it and suggest any improvements or alternatives.

PR #852, opened 2022-07-15
Pic 1971

rschallapalli:


vb-moj (approved):

src/main/java/uk/gov/justice/probation/courtcaseservice/service/OffenderUpdateService.java
```diff
@@ -36,10 +36,7 @@ public OffenderUpdateService(DefendantRepository defendantRepository, OffenderRe
     @Transactional
     public void removeDefendantOffenderAssociation(final String defendantId) {
         final var defendant = findDefendantOrElseThrow(defendantId);
-
-        if (!StringUtils.isEmpty(defendant.getCrn())) {
-            defendantRepository.save(defendant.withCrn(null).withId(null));
-        }
+        defendantRepository.save(defendant.withCrn(null).withManualUpdate(true).withOffenderConfirmed(true).withId(null));
```
jhevans, line 39: It won't actually affect this bug but we should add `.withManualUpdate(true).withOffenderConfirmed(true)` to line 65 as well so that the DB correctly reflects whether it's been confirmed/manually updated.

jhevans (changes requested):

jhevans (approved):

jhevans:
The pact tests are failing because of a different ticket which I did in the matcher. Feel free to merge and I'll follow up with a fix.

PR #873, opened 2022-08-09
PIC-2258

rschallapalli:


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/CaseCommentRequest.java
```diff
@@ -0,0 +1,30 @@
...
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE, force = true)
+@AllArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CaseCommentRequest {
+    @NotBlank
+    private final String caseId;
+    @NotBlank
+    private final String comment;
```
jhevans, line 22: I think we'll also need a field for the Author 🤔 We've got the created by fields which are populated from the user token but that only holds the username which often isn't very readable. This must be in scope in `prepare-a-case` for it to render it on the front end so we should be able to pass it through.

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CaseCommentEntity.java
```diff
@@ -0,0 +1,43 @@
...
+    @Column(name = "ID", updatable = false, nullable = false)
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @JsonIgnore
+    private final Long id;
+
+    @Column(name = "CASE_ID", updatable = false, nullable = false)
+    private final String caseId;
+
+    @Column(name = "COMMENT_ID", updatable = false, nullable = false)
+    private final String commentId;
```
jhevans, line 39: Why is there a separate commentId? Is this just so we don't expose our internal database keys?

src/main/java/uk/gov/justice/probation/courtcaseservice/service/CaseCommentsService.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private final CourtCaseRepository courtCaseRepository;
+    private final CaseCommentsRepository caseCommentsRepository;
+
+    public CaseCommentsService(CourtCaseRepository courtCaseRepository, CaseCommentsRepository caseCommentsRepository) {
+        this.courtCaseRepository = courtCaseRepository;
+        this.caseCommentsRepository = caseCommentsRepository;
+    }
+
+    public CaseCommentEntity createCaseComment(String caseId, CaseCommentEntity caseCommentEntity) {
```
jhevans, line 22: Could get the caseId from `caseCommentEntity` here and remove the first parameter

src/main/java/uk/gov/justice/probation/courtcaseservice/service/CaseCommentsService.java
```diff
@@ -0,0 +1,29 @@
...
+
+        CaseCommentEntity entity = caseCommentEntity.withCommentId(UUID.randomUUID().toString());
```
jhevans, line 24: Could use val 
```suggestion
       val entity = caseCommentEntity.withCommentId(UUID.randomUUID().toString());
```

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/model/CaseCommentRequestTest.java
```diff
@@ -0,0 +1,23 @@
...
+    @Test
+    void asEntity() {
+        var caseCommentRequest = CaseCommentRequest.builder()
+            .comment("awaiting PSR")
+            .caseId("test-cae-id")
+            .build();
+
+        assertThat(caseCommentRequest.asEntity())
+            .isEqualTo(CaseCommentEntity.builder()
+                .comment(caseCommentRequest.getComment())
```
jhevans, line 19: I'm not sure how I feel about using the getters from the `caseCommentRequest` here 🤔 It reads in a sort of circular manner because it refers back to the values of the test subject. It's also identical to the actual code in the `asEntity` method which doesn't feel right, as a reader it looks like it could have been copy-pasted.  

There's also an inconsequential typo on line 14 which the test didn't pick up. It doesn't really matter, it's still exercising the code, but it just doesn't look quite right to me.

jhevans (commented):
A few comments for review with one change I think we probably ought to do around explicitly passing in the `author`

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/CaseCommentRequest.java
rschallapalli: I thought created by is what we are interested in. Will update to receive the author.

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CaseCommentEntity.java
rschallapalli: That one too and it is not a good practice to use database generated PKs to use as resource ids. Though it should not matter in this case as we use case id access these always. Will update to use the id colum

src/main/java/uk/gov/justice/probation/courtcaseservice/service/CaseCommentsService.java
rschallapalli: Typo, will update with `var`

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/CaseCommentEntity.java
jhevans: I was just curious here, I wasn't suggesting you change it. You have a point about not exposing the database keys but wasn't sure what this was being used for. I suppose this getting a step ahead for when we need to start deleting them which is a good call.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/CaseCommentsService.java
jhevans: I actually asked myself which was right and got it wrong 😆 Confused by Kotlin 🙄

jhevans (dismissed):

src/main/resources/db/migration/courtcase/V2123__add_case_comments_table.sql
```diff
@@ -0,0 +1,21 @@
...
+create index case_id_idx on court_case (case_id);
+
+CREATE TABLE CASE_COMMENTS (
+   ID               SERIAL      PRIMARY KEY,
+   CASE_ID          TEXT        NOT NULL,
+   COMMENT          TEXT        NOT NULL,
+   AUTHOR           TEXT        NOT NULL,
+   CREATED          TIMESTAMP   NOT NULL DEFAULT now(),
+   CREATED_BY       TEXT        NULL,
+   CREATED_BY_UUID  TEXT        NOT NULL,
```
jhevans, line 12: Could use the `UUID` type here. In theory the same applies for CASE_ID but there's an [outstanding bug](https://dsdmoj.atlassian.net/browse/PIC-2375) which means this won't fly just yet

jhevans (commented):
One suggestion, would be good to use the UUID type

jhevans (approved):

PR #905, opened 2022-09-02
✅ PIC-2291: Add test for hmpps-events-topic health check

jhevans:


src/main/java/uk/gov/justice/probation/courtcaseservice/application/QueueConfig.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import uk.gov.justice.hmpps.sqs.HmppsQueueService;
+import uk.gov.justice.hmpps.sqs.HmppsTopic;
+
+import java.util.Optional;
+
+@Configuration
+public class QueueConfig {
```
vb-moj, line 11: Could this be TopicConfig or some generic class name to use for both the topic and queue instead?

vb-moj (approved):

PR #913, opened 2022-09-06
.PIC-2414

vb-moj:


jhevans (approved):
Good work, now we've established a solid pattern for testing our events it should make future work around them much easier 👍

PR #945, opened 2022-10-07
Pic 2481

vb-moj:


src/main/java/uk/gov/justice/probation/courtcaseservice/service/TelemetryService.java
```diff
@@ -191,6 +191,40 @@ public void trackOffenderProbationStatusUpdateEvent(OffenderEntity offenderEntit
...
+
+        ofNullable(offenderEntity)
+                .map(OffenderEntity::isPreSentenceActivity)
+                .ifPresent((preSentenceActivity) -> properties.put("preSentenceActivity", preSentenceActivity.toString()));
+
+        ofNullable(offenderEntity)
+                .map(OffenderEntity::getAwaitingPsr)
+                .ifPresent((awaitingPsr) -> properties.put("awaitingPsr", awaitingPsr.toString()));
+
+        addRequestProperties(properties);
```
vb-moj, line 223: Not sure we need to pass all the fields. May be just the CRN is enough?

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/ProbationStatusDetailComparator.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.justice.probation.courtcaseservice.service.model;
+
+import java.util.Comparator;
+
+import static java.util.Comparator.*;
+
+public class ProbationStatusDetailComparator {
```
rschallapalli, line 7: This looks much elegant using the language provision? 
public class ProbationStatusDetailComparator implements Comparator< ProbationStatusDetail> {
 ...
}

rschallapalli (changes requested):

src/main/java/uk/gov/justice/probation/courtcaseservice/service/model/ProbationStatusDetailComparator.java
vb-moj: Thanks,  On second thought, don't think we need a comparator at all.

src/main/java/uk/gov/justice/probation/courtcaseservice/service/TelemetryService.java
```diff
@@ -191,6 +191,40 @@ public void trackOffenderProbationStatusUpdateEvent(OffenderEntity offenderEntit
         telemetryClient.trackEvent(TelemetryEventType.OFFENDER_PROBATION_STATUS_UPDATED.eventName, properties, Collections.emptyMap());
     }
 
+
+    public void trackOffenderProbationStatusNotUpdateEvent(OffenderEntity offenderEntity) {
```
rschallapalli, line 195: AC doesn't mention this. Are these helpful? There could be thousands of these in a day. @jhevans

vb-moj: This is in the AC. `Probation status update which does not cause change does not trigger database call (log an event to indicate this)`

rschallapalli (approved):

PR #947, opened 2022-10-13
Pic 2436

vb-moj:


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/CourtCaseRequest.java
```diff
@@ -75,6 +75,8 @@ public class CourtCaseRequest {
     private final String nationality2;
     private final Boolean awaitingPsr;
     private final PhoneNumber phoneNumber;
+    private final String personId = "d1eefed2-04df-11ec-b2d8-0242ac130002";
```
jhevans, line 78: Don't think this assignment should be here 🤔

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/DefendantEntity.java
```diff
@@ -99,6 +100,9 @@ public class DefendantEntity extends BaseImmutableEntity implements Serializable
     @Column(columnDefinition = "jsonb", name = "PHONE_NUMBER")
     private final PhoneNumberEntity phoneNumber;
 
+    @Column(name = "PERSON_ID")
```
jhevans, line 103: Can this be made `nullable = false`?

src/main/resources/db/migration/courtcase/V2129__add_personid_to_defendant.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+    CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
jhevans, line 2: I don't believe this should be necessary as this extension should have been added in a previous migration.. it won't do any harm given the `IF NOT EXISTS` but I'm guessing you added it because you were having issues with it otherwise? 🤔 Are you aware of [this known issue](https://github.com/ministryofjustice/court-case-service#known-issues)? Not sure if that might be relevant?

src/main/resources/db/migration/courtcase/V2129__add_personid_to_defendant.sql
```diff
@@ -0,0 +1,7 @@
...
+    alter table DEFENDANT add column PERSON_ID UUID NULL;
+
+    update DEFENDANT SET PERSON_ID = uuid_generate_v4() where PERSON_ID is null;
+
```
jhevans, line 6: Could we follow up with an `ALTER` to add in a `NOT NULL` constraint once the `NULLS` have been populated?

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/Defendant.java
```diff
@@ -55,4 +57,8 @@ public String getSex() {
     @Valid
     @NotEmpty
     private final List<OffenceRequestResponse> offences;
+
+    public String getPersonId(){
+        return Optional.ofNullable(personId).orElse(UUID.randomUUID().toString());
```
rschallapalli, line 62: should the personId be assigned with the generated uuid before returning? Unlikely but prone to generating new uuid for every getPersonId call.
Also not sure the DTO should take this responsibility. I'd generate it outside of the DTO when needed if dto.getPersionId is null. Not precious about it but having a separate @Component to do generate the personId helps unit testing as mocking UUID is non trivial

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerPutByHearingIdIntTest.java
```diff
@@ -516,4 +521,39 @@ private List<PersonReferenceType> buildDefendantIdentifiers(String crn, String c
                 PersonReferenceType.builder().type("PNC").value(pnc).build()
         );
     }
+
+    @Test
+    void shouldPersistGivenPersonId_andCreateOneIfNotGiven_whenCreateOrUpdateHearingByHearingId() throws IOException {
+        //Person id given for defendant 1
+        String defendantId1 =   "1263de26-4a81-42d3-a798-bad802433318";
```
jhevans, line 528: Could use a var here

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseControllerPutByHearingIdIntTest.java
```diff
@@ -516,4 +521,39 @@ private List<PersonReferenceType> buildDefendantIdentifiers(String crn, String c
...
+        // known person id
+        defendantRepository.findFirstByDefendantIdOrderByIdDesc(defendantId1)
+                .ifPresentOrElse(defendantEntity -> {
+                    assertThat(defendantEntity.getPersonId()).isEqualTo(personIdForDefendant1);
+                }, () -> fail("Person id not matching"));
+
+        // person id unknown so check if exist
+        defendantRepository.findFirstByDefendantIdOrderByIdDesc(defendantId2)
+                .ifPresentOrElse(defendantEntity -> {
+                    assertThat(defendantEntity.getPersonId()).isNotBlank();
```
jhevans, line 556: You could go further here and use a [regex](https://github.com/ministryofjustice/court-case-matcher/blob/f245d82c4d163c007adb96d9b486ba23d61736ee/src/test/java/uk/gov/justice/probation/courtcasematcher/TestUtils.java) to confirm it's a valid UUID.

src/test/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedHearingRequestResponseTest.java
```diff
@@ -369,6 +374,47 @@ void givenOffencesWithJudicialResults_whenAsEntity_thenReturn() {
...
+                .defendantId(DEFENDANT_ID)
+                .build();
+        final var request = ExtendedHearingRequestResponse.builder()
+                .defendants(List.of(defendant))
+                .build();
+
+        final var hearingEntity = request.asHearingEntity();
+
+        assertThat(hearingEntity.getHearingDefendants()).hasSize(1);
+        assertThat(hearingEntity.getHearingDefendants().get(0).getDefendant().getPersonId()).isNotBlank();
```
jhevans, line 394: Again could use regex to confirm it's a UUID

jhevans (changes requested):
Looks good, I've left a few suggestions but the only critical one is a hard coded UUID in the `main` package which I'm guessing snuck in during development.

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/Defendant.java
vb-moj: As this is a temp solution until a proper person record service is built I have chosen this way as this will be easy to remove at a later stage when everything is ready.

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/CourtCaseRequest.java
vb-moj: @jhevans I am not sure why this class is in the main package. Not used anywhere except its own test class! I think we should remove this and its test class completely.

rschallapalli: True, can't see any usages for it. Should be removed.

src/test/java/uk/gov/justice/probation/courtcaseservice/testUtil/TestUtils.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.probation.courtcaseservice.testUtil;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class TestUtils {
+
+    public static final String UUID_REGEX = "[0-9a-fA-F]{8}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{12}";
+
+}
```
vb-moj, line 10: Thanks, @rschallapalli.

jhevans (approved):

PR #948, opened 2022-10-17
Pic 2435

vb-moj:


src/main/java/uk/gov/justice/probation/courtcaseservice/service/DomainEventService.java
```diff
@@ -76,7 +76,9 @@ private List<PersonReferenceType> buildDefendantIdentifiers(DefendantEntity defe
         return List.of(
                 PersonReferenceType.builder().type("CRN").value(defendant.getCrn()).build(),
                 PersonReferenceType.builder().type("CRO").value(defendant.getCro()).build(),
-                PersonReferenceType.builder().type("PNC").value(defendant.getPnc()).build()
+                PersonReferenceType.builder().type("PNC").value(defendant.getPnc()).build(),
+                PersonReferenceType.builder().type("PERSONID").value(defendant.getPersonId()).build()
```
rschallapalli, line 80: PERSON_ID?

vb-moj: I thought the same, but I believe this is a temp one until we have something as a type like CRN etc. Every system has to follow the same when receiving the event with the person id so I left it as it is. Happy to change if we all agree on something concrete.

rschallapalli: I'd follow the common practice of separating a multi word constant by underscore like we use camelcase for naming variables personId instead of persoinid

rschallapalli (approved):

PR #961, opened 2022-10-28
⛓ PIC-2542: Add confirmedOffender field to satisfy courtCaseMatcher Pact

jhevans:


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/mapper/CourtCaseResponseMapper.java
```diff
@@ -142,6 +142,7 @@ private static void addDefendantFields(CourtCaseResponseBuilder builder, Hearing
             .pnc(defendant.getPnc())
             .crn(hearingDefendantEntity.getCrn())
             .probationStatus(hearingDefendantEntity.getProbationStatusForDisplay())
+            .confirmedOffender(defendant.isOffenderConfirmed())
```
vb-moj, line 145: Just for clarification, How does the method name completely different from the field name `confirmedOffender?`  🤔

jhevans: The reason it's confirmedOffender rather than offenderConfirmed is just because of an oversight on my part when I was writing the matcher code, I didn't check that the field name would be consistent with the court-case-service field name. Given this is only needed temporarily I figured it wasn't worth re-doing all of the work in the matcher to get it to match.

vb-moj (approved):

PR #992, opened 2022-12-12
PIC-2613 clear up few more duplicates add unique constraint on heari…

rschallapalli:
…ng and defendant tables

### DIFF
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/DefendantRepositoryFacade.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/DefendantRepositoryFacade.java
@@ -14,6 +14,6 @@ public class DefendantRepositoryFacade {
     }
 
     public Optional<DefendantEntity> findFirstByDefendantId(String defendantId) {
-        return defendantRepository.findFirstByDefendantIdOrderByIdDesc(defendantId);
+        return defendantRepository.findFirstByDefendantId(defendantId);
     }
 }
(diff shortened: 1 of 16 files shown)


jhevans (approved):

vb-moj (dismissed):

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/DefendantRepositoryFacade.java
```diff
@@ -14,6 +14,6 @@ public DefendantRepositoryFacade(DefendantRepository defendantRepository) {
     }
 
     public Optional<DefendantEntity> findFirstByDefendantId(String defendantId) {
-        return defendantRepository.findFirstByDefendantIdOrderByIdDesc(defendantId);
+        return defendantRepository.findFirstByDefendantId(defendantId);
```
jhevans, line 17: In most of these cases I wonder whether we ought to be using `findByDefendantId` without the `first`, a) to make it clear that one result only is expected and b) to enforce that at the JPA as well as DB level

vb-moj: Agree with @jhevans

jhevans (approved):

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/DefendantRepositoryFacade.java
rschallapalli: It is already there in the code clean up ticket

PR #994, opened 2022-12-15
PIC-2662 Updated LocalStack docker compose config

cairnsp:
Updated the following to allow service to run locally:

- Updated docker-compose  config for LocalStack
- Update application-local.yml config for LocalStack
- Updated README file for running service locally

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ import java.time.LocalDate
 
 plugins {
     id 'java'
-    id 'org.springframework.boot' version '2.7.6'
+    id 'org.springframework.boot' version '2.7.5'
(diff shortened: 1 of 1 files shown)


vb-moj (dismissed):

build.gradle
```diff
@@ -3,7 +3,7 @@ import java.time.LocalDate
 
 plugins {
     id 'java'
-    id 'org.springframework.boot' version '2.7.5'
+    id 'org.springframework.boot' version '2.7.6'
```
vb-moj, line 6: I think we may need to revert it back to 2.7.5. I think we had some issues with this upgrade last time. Will see.

vb-moj (approved):

PR #1004, opened 2023-01-09
PIC-2679 Update CircleCi config to use hmpps-common-vars

cairnsp:


jhevans:
I've made a change that I think should fix the Jira integration 🤞

vb-moj (approved):

vb-moj (approved):

PR #1030, opened 2023-02-14
PIC-2660 Persist short term custody score

cairnsp:
- Enable calculation of short term custody score for incoming hearing data
- Persist score against each offence/defendant
- Update Offence Entity with score
- Flyway script for new column
- Enable Spring caching for CourtRepository


src/main/kotlin/uk/gov/justice/probation/courtcaseservice/service/ShortTermCustodyPredictorService.kt
```diff
@@ -5,33 +5,84 @@ import hex.genmodel.easy.RowData
...
+                        val score = calculateShortTermCustodyPredictorScore(
+                            PredictorParameters(
+                                it.name,
+                                calculateAge(defendant),
+                                homeOfficeOffenceCode
+                            )
+                        )
+                        offence.shortTermCustodyPredictorScore = BigDecimal.valueOf(score)
+                    }, {
+                        log.warn("No court name could be found for court code: ${hearingEntity.hearingDays[0].courtCode}")
```
vb-moj, line 78: Nothing wrong here but could this be moved into a separate function instead? From try/catch block onwards.

vb-moj (approved):

PR #1038, opened 2023-02-23
PIC-2735

vb-moj:
. Added result text to judicial result entity.
. Updated unit tests.
. Updated IT to verify the result text is persisted.


vb-moj:
> Is this request only to persist resultText, don't we have to return it from a call to `GET /hearing/{hearingId}`

I think it should, I thought I have updated that as well, will check. Thanks

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedHearingRequestResponse.java
```diff
@@ -256,6 +257,7 @@ private List<JudicialResultEntity> buildJudicialResults(List<JudicialResult> jud
                             .isConvictedResult(judicialResult.isConvictedResult())
                             .label(judicialResult.getLabel())
                             .judicialResultTypeId(judicialResult.getJudicialResultTypeId())
+                            .resultText(judicialResult.getResultText())
```
vb-moj, line 260: @rschallapalli  This change will return the result text for the GET.

rschallapalli: Yes, right, my bad

rschallapalli (approved):

PR #1051, opened 2023-03-21
PIC-2733

vb-moj:
. Added script to create case marker entity.
. Added case marker entity to hearing.
. Updated tests.

### DIFF
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/Plea.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/Plea.java
@@ -3,6 +3,7 @@ package uk.gov.justice.probation.courtcaseservice.controller.model;
 
 import io.swagger.v3.oas.annotations.media.Schema;
 import lombok.*;
+import uk.gov.justice.probation.courtcaseservice.jpa.entity.PleaEntity;
 
 import java.time.LocalDate;
 
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/Plea.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/Plea.java
@@ -2,8 +2,11 @@ package uk.gov.justice.probation.courtcaseservice.controller.model;
 
 
 import io.swagger.v3.oas.annotations.media.Schema;
-import lombok.*;
-import uk.gov.justice.probation.courtcaseservice.jpa.entity.PleaEntity;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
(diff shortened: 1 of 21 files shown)


src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/HearingEntity.java
rschallapalli, line 1: Can we check if the case markers should be associated with CourtCase instead of Hearing?
If a new hearing comes and there are no case markers are we not supposed to show the case markers when the user view the summary?  
The implementation might be right but just want to clarify.

rschallapalli (changes requested):

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/entity/HearingEntity.java
vb-moj: Good point but I am not sure. Will check with Tony.

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/Plea.java
```diff
@@ -3,6 +3,7 @@
 
 import io.swagger.v3.oas.annotations.media.Schema;
 import lombok.*;
+import uk.gov.justice.probation.courtcaseservice.jpa.entity.PleaEntity;
```
rschallapalli, line 6: Unused import?

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/Verdict.java
```diff
@@ -7,6 +7,7 @@
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import uk.gov.justice.probation.courtcaseservice.jpa.entity.VerdictEntity;
```
rschallapalli, line 10: Unused import?

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
```diff
@@ -146,6 +146,7 @@ private Mono<HearingEntity> createOrUpdateHearing(String hearingId, final Hearin
 
     private static void addHearingToCase(HearingEntity updatedHearing, CourtCaseEntity courtCaseEntity) {
         courtCaseEntity.addHearing(updatedHearing);
+        //courtCaseEntity.addCaseMarkers(updatedHearing.getCourtCase().getCaseMarkers());
```
rschallapalli, line 149: Can be removed?

rschallapalli (approved):

PR #1109, opened 2023-09-05
PIC-3291 Force HMPPS SQS to use web token

cairnsp:
Assumes you will be using Web Identity Token credentials and thus won't need any queue / topic access keys or secrets


rschallapalli (approved):
LGTM

markberridge (approved):
LGTM

PR #1111, opened 2023-09-07
PIC-3280 Increase Java heap size

cairnsp:
We've increased k8s pod resources in the past, but didn't up the Java heap size accordingly.

I've increased both the heap size and the k8s resources size to realistic values for production spring boot services.


markberridge:
That makes sense!

vb-moj (approved):

PR #1116, opened 2023-09-11
PIC-3183 add resulted datetime when case is resulted and return in ou…

rschallapalli:
…tcomes response


johnyfarrar (dismissed):
LGTM

johnyfarrar (approved):
All looks fine!

PR #1139, opened 2023-11-28
PIC-3499 create missing hibernate envers index in prod

rschallapalli:


paddybasi (dismissed):
LGTM

rschallapalli:
> I am wondering why this is something that needs to be done manually, shouldn't the migration framework handle/support this? Did a bit of digging and this came up https://stackoverflow.com/questions/26221688/hibernate-envers-revinfo-table-doesnt-exist

Those configurations have caused more problems and were also major reason for current instability of the service (Accidentally, leaving in framework's hands to create/update the tables.)

johnyfarrar (approved):
LGTM

PR #1161, opened 2024-03-04
Pic 3588 doc upload api

rschallapalli:


src/main/kotlin/uk/gov/justice/probation/courtcaseservice/client/HmppsDocumentManagementApiClient.kt
```diff
@@ -0,0 +1,115 @@
...
+                handle5xxError("${it.statusCode()} error uploading document",
+                    HttpMethod.POST, documentPath, ExternalService.DOCUMENT_MANAGEMENT_API)
+            }
+            .bodyToMono(DocumentUploadResponse::class.java)
+            .doOnSuccess {
+                log.info("Document upload success {}", it)
+            }
+    }
+
+    fun getDocument(documentUuid: String): ResponseEntity<Flux<InputStreamResource>>? {
```
AndySweilam, line 60: Use flows/co-routines and avoid blocking in client.
Start here:
fun getDocument(documentUuid: String): Flow<InputStreamResource>

AndySweilam (approved):
The code uses fluxes, we moved away from these. This will need to be changed to o-routines. 
File type checks also needed. @andrewcharnley is aware of this.

PR #1179, opened 2024-03-21
Update Spring Boot to accept max 50MB files

johnyfarrar:


markreesmoj (approved):
Seems OK. It would be nice to see an integration test for this which starts a server and tries to upload a file which is over the default limit  - [Apparently 1MB](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/web/servlet/MultipartProperties.html)

Although I can quite see why you have not done so - there doesn't seem to be an existing test which could easily be enhanced.

[This is an interesting discussion](https://stackoverflow.com/questions/33667276/test-maximum-upload-file-size-with-mockmultipartfile) on how to do it - as so often with Stack Overflow these days the accepted answer with the most upvotes is incorrect!

PR #1182, opened 2024-03-25
Increase the Nginx/ingress body size limit for requests for DEV

Obsiye:
### Increase the Nginx/ingress body size limit for requests for DEV

This is because prepare a case service is getting a 413 http response
"client intended to send too large body: 6156882 bytes" as it tries to upload a 6mb file.

The Cloud Platform Kibana Ingress Logs suggest it's due to the `court-case-service-proxy` ingress server defaulting to a max of 1megabyte

**Additional information:**

https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#custom-max-body-size


johnyfarrar (approved):
LGTM

PR #1183, opened 2024-03-25
Increase client_max_body_size in nginx/ingress

Obsiye:
Increase the file upload size limit to 50M by updating the nginx.conf file `client_max_body_size` value and also updating the ingress.yaml

This is to allow a large file upload from prepare-a-case service. The default nginx limit is 1mb which isn't enough.

Undo this previous PR as we have an nginx.conf file available: https://github.com/ministryofjustice/court-case-service/pull/1182


johnyfarrar (approved):
LGTM

johnyfarrar (approved):
LGTM

PR #1186, opened 2024-03-27
Fix Legacy property on case comments

johnyfarrar:


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/CaseCommentResponse.java
```diff
@@ -24,17 +24,19 @@ public class CaseCommentResponse {
...
-            .build();
+                .comment(caseCommentEntity.getComment())
+                .commentId(caseCommentEntity.getId())
+                .caseId(caseCommentEntity.getCaseId())
+                .defendantId(caseCommentEntity.getDefendantId())
+                .created(caseCommentEntity.getCreated())
+                .author(caseCommentEntity.getAuthor())
+                .createdByUuid(caseCommentEntity.getCreatedByUuid())
+                .draft(caseCommentEntity.isDraft())
+                .legacy(caseCommentEntity.isLegacy())
```
Obsiye, line 39: I think if you want to use `isLegacy()` you'd need to change the column name to `IS_LEGACY` in the `CaseCommentEntity.java`

like the way it does for "IS_DRAFT"
```
 @Column(name = "IS_DRAFT", nullable = false)
    private boolean draft;
 ```
 
 or you can use `caseCommentEntity.getLegacy()` i think

johnyfarrar: It autocompletes to isLegacy based on the boolean

Another example here - 

Screenshot 2024-03-27 at 15 59 03

Obsiye: ah yes, sorry...Lombok 👍🏾

PR #1205, opened 2024-04-18
Allow Null URN value in HmppsDocumentApiMetadata

Obsiye:
#### Allow Null URN value in HmppsDocumentApiMetadata

This is because the `court_case urn` has some null values in the database. Therefore some document uploads are throwing null pointer exceptions because its `court_case#getUrn()` is null

- Add unit test for the `HmppsDocumentApiMetadata` data model
- Update integration test to test that the case urn has been stored


andrewcharnley (approved):
LGTM

johnyfarrar (approved):
LGTM

PR #1212, opened 2024-05-08
Revert assigned users bug fix release

Obsiye:
**Revert:**

[Return all assigned users and create v2 endpoint](https://github.com/ministryofjustice/court-case-service/pull/1194)
[Remove duplicates after fixing assigned users bug](https://github.com/ministryofjustice/court-case-service/pull/1210)

**DescriptIon:**

Revert the assigned users bug fix release, after the UI had timeout issues when calling the hearing-outcomes endpoint. This is due to the way we've altered handling of pagination - away from the DB and into the code. 

If we plan to reintroduce this fix, we must do some performance profiling and find ways to reduce the time taken to get all the hearing-outcomes data


johnyfarrar (approved):
LGTM

PR #1235, opened 2024-06-20
Revert "Pic 3904"

Obsiye:
Reverts ministryofjustice/court-case-service#1233

It was initially merged for us devs to do testing. also it's because we can test on the event driven architecture on DEV


johnyfarrar (approved):
LGTM

PR #1237, opened 2024-06-25
PIC-3904: The PUT req for assigning hearinig outcome to the current u…

jpanesar07:
…ser was taking too long. Only initialize the objects needed


johnyfarrar (approved):
LGTM

PR #1248, opened 2024-07-15
PIC-3939 Add index on defendant names

Obsiye:
Add an index to find a defendant by PNC, forename1 and surname

We have an SQL taking up a lot of CPU and making many calls:
```
select * from defendant where pnc= $1 and date_of_birth = $2 and lower(name->>?) = lower($3) and lower(name->>?) = lower($4)
```
This index should hopefully reduce the average speed and CPU usage


jpanesar07 (approved):

johnyfarrar (approved):
LGTM

PR #1249, opened 2024-07-17
PIC-4018 Set timestamp threshold for processing messages

Obsiye:
Set a threshold of 2 days

Do not process messages that are older than 2 days for the `picnewoffendereventsqueue`. As it has already been processed.

This should reduce the number of queries made to our database and therefore improve performance.


jpanesar07 (approved):

johnyfarrar (approved):
LGTM

PR #1267, opened 2024-08-21
PIC-4022 BugFix: Return latest draft comment from list of draft comments

Obsiye:
Due to a data race, more than 1 draft comments can be saved in the database which shouldn't happen.

This leads to users unable to save a comment, which is caused by the below `NonUniqueResultException` exception

Most commonly, more than 1 draft comment exists that were created milliseconds apart.
`jakarta.persistence.NonUniqueResultException: query did not return a unique result: 2`

This fix returns and updates only the latest draft comment and deletes the older draft comments.

Also, add a unit test


johnyfarrar (approved):
Looks good to me!

PR #1275, opened 2024-09-16
PIC-4154: Ability to filter case list data on multiple criteria, for …

jpanesar07:
…the matcher scenario


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
```diff
@@ -1,71 +1,35 @@
...
-import uk.gov.justice.probation.courtcaseservice.service.AuthenticationHelper;
-import uk.gov.justice.probation.courtcaseservice.service.CaseCommentsService;
-import uk.gov.justice.probation.courtcaseservice.service.CaseProgressService;
-import uk.gov.justice.probation.courtcaseservice.service.CourtCaseService;
-import uk.gov.justice.probation.courtcaseservice.service.HearingNotesService;
-import uk.gov.justice.probation.courtcaseservice.service.OffenderMatchService;
-import uk.gov.justice.probation.courtcaseservice.service.OffenderUpdateService;
+import uk.gov.justice.probation.courtcaseservice.controller.model.*;
+import uk.gov.justice.probation.courtcaseservice.jpa.entity.*;
+import uk.gov.justice.probation.courtcaseservice.service.*;
```
markreesmoj, line 24: Seems a pity to introduce star imports where there weren't any previously

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
```diff
@@ -359,6 +323,55 @@ public ResponseEntity<CaseListResponse> getCaseList(
...
+        final var hearingSearchFilter = HearingSearchFilter.builder()
+                .courtCode(courtCode)
+                .hearingDay(date)
+                .source(null)
+                .breach(false)
+                .build();
+
+        var courtCases = courtCaseService.filterHearings(hearingSearchFilter);
+
+        var courtCaseResponses = filterCourtCaseResponses(date, numberOfPossibleMatches, forename, surname, defendantName, caseId, hearingId, defendantId, courtCases);
```
markreesmoj, line 370: The previous block of code builds up a filter object to pass into the filter method, rather than passing in a lot of arguments - might be more readable to do that here too?

It also delegates the filtering to another class rather than doing it in the controller

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
jpanesar07: Corrected

src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java
jpanesar07: Agreed, moved the logic down to the service and used the filter wrapper

Obsiye (approved):
Looks good to me 🎉

PR #1279, opened 2024-09-24
PIC-4138 Remove logs after resolving bug

Obsiye:
Remove logs that were used solely to debug a null pointer exception. It has been resolved so there's no longer a need for the logs.


johnyfarrar (approved):
LGtM

PR #1286, opened 2024-10-02
PIC-4207: Return lastUpdate date/time as part of the response from /h…

jpanesar07:
Return lastUpdate date/time as part of the response from /hearing/{hearingId}

### DIFF
--- a/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedHearingRequestResponse.java
+++ b/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedHearingRequestResponse.java
@@ -63,6 +63,8 @@ public class ExtendedHearingRequestResponse {
     private final String listNo;
     private final List<CaseMarker> caseMarkers;
 
+    private final LocalDateTime lastUpdated;
+
     public static ExtendedHearingRequestResponse of(HearingEntity hearing) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/probation/courtcaseservice/controller/model/ExtendedHearingRequestResponse.java
```diff
@@ -63,6 +63,8 @@ public class ExtendedHearingRequestResponse {
     private final String listNo;
     private final List<CaseMarker> caseMarkers;
 
+    private final LocalDateTime lastUpdated;
+
     public static ExtendedHearingRequestResponse of(HearingEntity hearing) {
```
markreesmoj, line 68: This method is also called by the PUT mapping which creates/updates hearings at 
https://github.com/ministryofjustice/court-case-service/blob/ffd4756a649f63866ee77eaee2076bafa6cd6f61/src/main/java/uk/gov/justice/probation/courtcaseservice/controller/CourtCaseController.java#L132

Are we confident that it will not have any effect on that process?

jpanesar07: Yes, its just used as part of the response in the put endpoint as well, so I can't see it being any different to the get

PR #1296, opened 2024-11-01
Add read only database user for developers

Obsiye:
Add a read only database user for developers to use when accessing the database

This is so that developers can connect to the database with read only permissions and it prevents any accidental writes

### DIFF
--- a/helm_deploy/court-case-service/templates/_envs.tpl
+++ b/helm_deploy/court-case-service/templates/_envs.tpl
@@ -193,6 +193,18 @@ env:
         name: sqs-pic-new-offender-events-dlq-secret
         key: sqs_queue_name
 
+  - name: DATABASE_READONLY_USERNAME
+    valueFrom:
+      secretKeyRef:
+        name: court-case-service-readonly-user
+        key: database_readonly_username
+
+  - name: DATABASE_READONLY_PASSWORD
+    valueFrom:
+      secretKeyRef:
+        name: court-case-service-readonly-user
+        key: database_readonly_password
+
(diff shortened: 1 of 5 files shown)


src/main/resources/application-local.yml
```diff
@@ -77,6 +77,11 @@ spring:
   flyway:
     defaultSchema: courtcaseservice
     locations: classpath:db/migration/courtcase, filesystem:src/test/resources/db/migration/local, filesystem:src/test/resources
+    placeholders:
```
markreesmoj, line 80: Does this belong under `flyway`?

helm_deploy/court-case-service/templates/_envs.tpl
```diff
@@ -193,6 +193,18 @@ env:
...
+      secretKeyRef:
+        name: court-case-service-readonly-user
+        key: database_readonly_username
+
+  - name: DATABASE_READONLY_PASSWORD
+    valueFrom:
+      secretKeyRef:
+        name: court-case-service-readonly-user
+        key: database_readonly_password
+
```
markreesmoj, line 207: I am not sure that we would want to expose this to the application itself. I appreciate it is for the purposes of running the flyway script but I can't see any further use for it. 

Have you thought about using terraform for this? It seems like it might be generally useful to all teams.

https://stackoverflow.com/questions/60820805/create-read-only-postgres-user-with-terraform is a starting point

Obsiye:
Requested cloud platform handle this and tehy suggested a read replica which is a bit more than what we need

PR #1313, opened 2024-12-03
Pic 4372

jpanesar07:


helm_deploy/court-case-service/templates/ingress.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
     # Secure the retry all dlqs endpoint from outside of the Kubernetes ingress
     nginx.ingress.kubernetes.io/server-snippet: |
       server_tokens off;
-      location ~* ^/(queue-admin/retry-all-dlqs|process-un-resulted-cases) {
+      location ~* ^/(queue-admin/retry-all-dlqs|process-un-resulted-cases|hearing/delete-duplicates) {
```
markreesmoj, line 24: strange that this does not have a leading slash but as long as it works then it is fine

src/main/resources/application.yml
```diff
@@ -201,6 +201,7 @@ offender-service:
 feature:
   flags:
     enable-cacheable-case-list: true
+    delete-hearing: true
```
markreesmoj, line 204: Don't we just want to set this to `true` in preprod and prod when reqjuired? I don't see the point in defining it here if we are overriding it in all environments

src/main/java/uk/gov/justice/probation/courtcaseservice/service/ImmutableCourtCaseService.java
```diff
@@ -162,7 +162,7 @@ public List<HearingEntity> filterHearings(HearingSearchFilter hearingSearchFilte
     @Override
     public Optional<CourtCaseEntity> findByCaseId(String caseId) {
         assert courtCaseRepository != null;
```
markreesmoj, line 164: !!

src/main/java/uk/gov/justice/probation/courtcaseservice/service/DeleteHearingsService.java
```diff
@@ -0,0 +1,46 @@
...
+@Slf4j
+@Service
+public class DeleteHearingsService {
+
+    private final HearingRepositoryFacade hearingRepositoryFacade;
+    private final DuplicateHearingRepository duplicateHearingRepository;
+    private final TelemetryService telemetryService;
+
+    @Autowired
+    private FeatureFlags featureFlags;
```
markreesmoj, line 24: why isn't this in the constructor?

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/HearingRepository.java
```diff
@@ -91,7 +91,8 @@ List<HearingEntity> findByCourtCodeAndHearingDay(
     @Query(value = "select * from hearing where id in (" +
         "select max(h.id) from hearing h, court_case cc where " +
         "cc.case_id = :caseId " +
-        "and h.fk_court_case_id = cc.id group by h.hearing_id)",
+        "and h.deleted = false " +
```
markreesmoj, line 94: do we need to add this line to all these native queries? 😱

src/main/java/uk/gov/justice/probation/courtcaseservice/service/DeleteHearingsService.java
markreesmoj: I can see now - it is being modified in the IntegrationTest post construction

src/main/java/uk/gov/justice/probation/courtcaseservice/jpa/repository/HearingRepository.java
jpanesar07: I know, only that one so far.

markreesmoj (approved):
I will have a proper look later on but for now let's give this a try in non-production environments

markreesmoj (commented):
💬

src/main/java/uk/gov/justice/probation/courtcaseservice/application/FeatureFlags.java
```diff
@@ -2,9 +2,12 @@
 
 import java.util.HashMap;
 import java.util.Map;
+
+import lombok.Getter;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.stereotype.Component;
 
+@Getter
```
markreesmoj, line 10: why is this necessary?

PR #1321, opened 2025-01-07
PIC-4373: Ensure deleted hearing outcomes are filtered out.

jpanesar07:
Ensure deleted hearing outcomes are filtered out before returning data from the api


johnyfarrar (approved):
LGTM

PR #1326, opened 2025-01-30
Revert "Update dependency uk.gov.justice.service.hmpps:hmpps-kotlin-spring-boot-starter to v1"

Obsiye:
Reverts ministryofjustice/court-case-service#1221


Obsiye:
Revert this because we're getting the error
`NoSuchMethodError: 'void org.springdoc.core.service.OpenAPIService.setServerBaseUrl(java.lang.String)'`

when trying to access swagger. Other than that, the service looks to be working good 🤔

johnyfarrar (approved):
LGTM

PR #1340, opened 2025-02-10
BugFix nomis auth url append /auth

Obsiye:
Append /auth to the nomis auth url ... after work done on the Pinger class


KiraNamida (approved):
LGTM :shipit:

PR #1349, opened 2025-02-24
Revert "Add uniq constraint on case_defendant table"

Obsiye:
Reverts ministryofjustice/court-case-service#1344

This is because the uniq index cannot be created when there are already duplicates in the table


Obsiye:
No need to revert, we fix-forwarded

PR #1352, opened 2025-02-25
Use identity column for primary keys

Obsiye:
Hopefully this should resolve `ERROR: duplicate key value violates unique constraint \"court_case_new_pkey\` exceptions

This is because the court_case table has no sequence and no identity column


Obsiye:
It's an identity column already. I'm assuming it's to do with concurrent table insertions affecting the implied database sequence for the id column in court_case table

PR #1378, opened 2025-04-24
PIC-4572: Adding cprUUID to defendant entity and corresponding objects

jpanesar07:


src/main/resources/db/migration/courtcase/V2193__add_cpr_uuid_to_defendant.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+    alter table if exists defendant ADD COLUMN cpr_uuid TEXT;
```
Obsiye, line 2: could it have the type `UUID`?

jpanesar07: Yes, updated

Obsiye (approved):

PR #1394, opened 2025-05-23
PIC-4650 update to 8.2.0 that fixes cves too

Obsiye:
This should fix this CVE for us https://spring.io/security/cve-2025-22233

This should also fix our other CVE


markreesmoj (approved):
👍

PR #1499, opened 2026-05-18
ignore outcome not required for auto outcomes

RichT900:


leesykes (approved):
I assume you've grep-ed around to make sure this query only needs changing in these two locations. And is there a test we could update to avoid regressions? I was just feeling pedantic today.

PR #1501, opened 2026-05-21
PIC-5402 - implemented tests for SAR

michaelpitt1:


src/test/resources/application-test.yml
```diff
@@ -46,6 +46,9 @@ flipt:
   api-key: "test-api-key"
 
 spring:
+  autoconfigure:
+    exclude:
+      - uk.gov.justice.hmpps.test.kotlin.auth.JwtAuthorisationHelper
```
Aron-v1, line 51: Does this affect the other tests?

docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
     environment:
       POSTGRES_USER: root
       POSTGRES_PASSWORD: dev
+    command: postgres -c max_connections=200
```
Aron-v1, line 12: Is this related to the SAR change?

Aron-v1 (approved):
LGTM, 2 queries to double check 👍

src/test/resources/application-test.yml
michaelpitt1: Thanks Aron. I had trouble with conflicting tests but I managed to resolve it this way.

docker-compose.yml
michaelpitt1: yes, I added this because the new integration tests were exceeding the default max connections of just 10.

